PR #11, opened 2021-02-16
Chore | Add project devDependencies (eslint, prettier, lint-staged, commitlint)

richiemccoll:
This change adds some more project configuration.

- `eslint` & `prettier` (maybe some of the lint settings could be tuned further?)
- `lint-staged` to run on `pre-commit` 
- `commitlint` to ensure that commit messages are consistent


components/Auth.js
```diff
@@ -23,7 +23,8 @@ WebBrowser.maybeCompleteAuthSession()
 export default function Auth() {
   const [user, setUser] = useState()
 
-  const [request, response, promptAsync] = Google.useIdTokenAuthRequest({
+  // eslint-disable-next-line no-unused-vars
+  const [_, response, promptAsync] = Google.useIdTokenAuthRequest({
```
simoneb, line 27:
```suggestion
  const [, response, promptAsync] = Google.useIdTokenAuthRequest({
```

simoneb (approved):
LGTM. You may want to consider using husky 5, which was tagged as latest version a few days ago, but it requires a different syntax for declaring the hooks.

components/Auth.js
richiemccoll: Fixed https://github.com/nearform/optic-expo/pull/11/commits/a304d26daff302384ae7bdaba7c933bff89b5fc8.

PR #13, opened 2021-02-16
Feat | Add secret manually

### LINKED ISSUE #3: add new secret manually

n370:
Closes #3.

Major changes:

- [x] Add react-navigation and re-structure the app little bit to fit the library patterns.
- [ ] Port over secret persistence manager from the web app here. (Still needs work, see Simone's comments [here](https://github.com/nearform/optic-expo/pull/13#discussion_r577961708) and [here](https://github.com/nearform/optic-expo/pull/13#discussion_r577962993))
- [x] Add a react-navigation screen and a screen component for the "type to add secret" feature.
- [x] Add placeholders for the screens and screen component placeholders for the other "add secret" strategies.

What is not included here:

- Adding UI components for the secrets list and the items it holds

I ended up moving things around and that partly explains the many files changed. I'm sorry about that.


simoneb (commented):
I see lots of code that is in the other open PR, not sure if that's on purpose. If your change depends on that other PR, please use that branch as the base so those changes don't appear in here.

Please see the review of that PR for requested changes which are primarily around conventions and apply them here too.

Main.js
```diff
@@ -1,24 +1,60 @@
...
+      title: 'Optic',
+    },
+    type: {
+      title: 'New Secret',
+    },
+  },
+}
 
 export default function Main() {
+  const [secrets, setSecrets] = useState([])
```
simoneb, line 25: I now see what you meant with your question on slack. The secrets should not be in local state unless there's a good reason why we want to store them there. React Native has native storage that we can use, similarly to how we use localStorage in the Web app.

Main.js
```diff
@@ -1,24 +1,60 @@
...
+            options={{ title: UI_STRINGS.routes.home.title }}
+          >
+            {props => <Home {...props} user={user} secrets={secrets} />}
+          </Stack.Screen>
+          <Stack.Screen
+            name="Type"
+            options={{ title: UI_STRINGS.routes.type.title }}
+          >
+            {props => (
+              <TypeNewSecret {...props} user={user} addSecret={addSecret} />
```
simoneb, line 49: the user object can be grabbed from React context, let's not pass it around as props so we can simplify the way we declare these screens. We can do without passing any props to the components

Main.js
```diff
@@ -1,24 +1,60 @@
...
+            )}
+          </Stack.Screen>
+        </Stack.Navigator>
+      ) : (
+        <Stack.Navigator>
+          <Stack.Screen name="Auth" component={Auth} />
+        </Stack.Navigator>
```
simoneb, line 56: I don't think you need a stack for the login screen

lib/secrets.js
```diff
@@ -0,0 +1,77 @@
+let DUMMY_UIID_GEN = 0
+let DUMMY_STORE = []
```
simoneb, line 2: https://reactnative.dev/docs/asyncstorage

n370:
Hey @simoneb, thanks for your comments. I just now had a chance to look at them, I guess I addressed a few with the latest changes I pushed a moment ago. I'm opening this up for review and updating the description with the things I'd like reviewers to pay attention to as well as known issues and trade-offs to disregard for now.

Main.js
n370: I'm taking a stab at this tomorrow first thing unless you think it is better if merge this one as is and improve upon  it on a different PR. I have moved things around a little bit so it may be worth to merge as is to easy a little bit possible conflicting changes with @richiemccoll and others. Whatever y'all think is best.

components/Auth.js
```diff
@@ -3,20 +3,17 @@ import * as WebBrowser from 'expo-web-browser'
...
 
-import { useAuthenticationContext } from '../context/authentication'
-import theme from '../defaultTheme'
+import theme from '../lib/defaultTheme'
 
 import Spacer from './Spacer'
 
 const UI_STRINGS = {
-  title: 'Optic',
+  heading: 'Optic',
```
n370, line 11: I'm rolling this one back, please disregard.

components/Main.js
```diff
@@ -0,0 +1,84 @@
...
+      title: 'New Secret',
+    },
+  },
+}
+
+export default function Main() {
+  return (
+    <NavigationContainer>
+      <Stack.Navigator
+        {...{
```
simoneb, line 37: spreading not needed

components/TypeNewSecret.js
```diff
@@ -0,0 +1,84 @@
...
+
+export default function TypeNewSecret() {
+  const { user } = useAuthenticationContext()
+  const { add } = useSecretsContext()
+  const [issuer, setIssuer] = useState('')
+  const [secret, setSecret] = useState('')
+  const [account, setAccount] = useState(user.displayName)
+  const { navigate } = useNavigation()
+
+  const handleOnPress = async () => {
```
simoneb, line 19: naming: this handler could use a more meaningful name

components/TypeNewSecret.js
```diff
@@ -0,0 +1,84 @@
...
+    </View>
+  )
+}
+
+const UI_STRINGS = {
+  issuerTextInputLabel: 'Issuer',
+  secretTextInputLabel: 'Secret',
+  accountTextInput: 'Account',
+  addSecretButtonLabel: 'Add Secret',
+}
```
simoneb, line 66: lets' try to be consistent with the positioning of the ui strings, as we do with the styles. either above or below component declarations

simoneb (approved):
LGTM. Two notes:

- I pushed a commit because a change in this PR had inadvertently reintroduced the login issue (please always test your changes on a device)
- there are two warnings pertaining the inline components currently used for two screens. Not a big deal as they're temporary though

richiemccoll (approved):
LGTM. We should remember to add some tests for this UI after [this](https://github.com/nearform/optic-expo/pull/20) has been merged

Issue #16, opened 2021-02-17
Setup Jest and test infrastructure


richiemccoll:
- Jest with `jest-expo` setup
- `react-native-testing-library`
- Add a new CI step for tests
- Add tests for Auth / Home screens.

PR #20, opened 2021-02-18
feat: setup jest and testing library/native

richiemccoll:
Closes https://github.com/nearform/optic-expo/issues/16

This PR adds a few new devDependencies.
- `@testing-library/jest-native`
- `@testing-library/react-native`
- `jest` & `jest-expo`

I've added some super basic tests that check correct text is rendered on screen and that users can log in. I've also updated the Github actions to now run `yarn test` as well.

Screenshot 2021-02-18 at 10 03 54


simoneb (approved):
LGTM, nice

n370 (approved):

PR #21, opened 2021-02-18
feat: update auth screen ui

richiemccoll:
This PR adds:
- Some fonts that the designs are using. `Poppins_700Bold` and `DidactGothic_400Regular`. 
- The `NearForm` logo (exported from figma)
- `Headline` and `BodyText` typography styles from the design system in Figma.
- A new dependency `expo-app-loading` - can be used for the splash screen (not included in this PR)

~Note: The only thing I couldn't get working was the color of the Google icon. The Button from react-native-paper was setting the color of the icon to black. Maybe someone has a better idea on how to override that.~ Done here https://github.com/nearform/optic-expo/pull/21/commits/9477af4da284bbc74add4267e54bc65ec1a97330


App.js
```diff
@@ -1,13 +1,29 @@
...
 import theme from './lib/defaultTheme'
 
 export default function App() {
+  usePoppins({
+    Poppins_700Bold,
+  })
+
+  useDidactGothic({
+    DidactGothic_400Regular,
+  })
```
simoneb, line 25: you have to wait until the fonts are loaded. see the docs

components/Auth.js
```diff
@@ -28,21 +30,23 @@ export default function Auth({ handleLogin }) {
...
+      <Spacer size={12} />
+      <Headline color={theme.colors.surface}>{UI_STRINGS.heading}</Headline>
+      <Spacer size={4} />
+      <BodyText style={styles.subheading}>{UI_STRINGS.subheading}</BodyText>
+      <Spacer size={12} />
+      <Button
+        style={styles.button}
+        accessibilityLabel="login"
+        mode="contained"
+        icon="google"
```
simoneb, line 42: I believe the icon can also be an image. see the paper docs. using the string icon name will give you [this](https://materialdesignicons.com/icon/google), whose color can't be customized as in the designs

components/typography.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { Text } from 'react-native-paper'
+import { StyleSheet } from 'react-native'
+
+export function Headline({ children, color, style, ...props }) {
+  return (
+    <Text
+      style={[styles.headline, color && { color }, color && { color }, style]}
```
simoneb, line 8:
```suggestion
      style={[styles.headline, color && { color }, style]}
```

components/typography.js
```diff
@@ -0,0 +1,39 @@
...
+    >
+      {children}
+    </Text>
+  )
+}
+
+export function BodyText({ children, color, style, ...props }) {
+  return (
+    <Text
+      style={[styles.bodyText, color && { color }, color && { color }, style]}
```
simoneb, line 19:
```suggestion
      style={[styles.bodyText, color && { color }, style]}
```

components/typography.js
```diff
@@ -0,0 +1,39 @@
...
+    fontWeight: '700',
+    fontSize: 60,
+  },
+  bodyText: {
+    fontFamily: 'DidactGothic_400Regular',
+    fontWeight: '400',
+    fontSize: 16,
+    lineHeight: 24,
+  },
+})
```
simoneb, line 39: important thing to note. if we need custom fonts, we can't use RN's Text anymore. We must use either a custom set of typograpy components (as you're doing here), or change the font configuration of paper's typography components to use our fonts. see https://callstack.github.io/react-native-paper/theming.html and the source code of paper's default theme to see how font customization is done https://github.com/callstack/react-native-paper/blob/main/src/styles/DefaultTheme.tsx

simoneb (changes requested):

components/typography.js
richiemccoll: Do you think it's worth me sticking that in a comment in this file?

simoneb: sure. my comment is not exactly accurate though, it's not true that we can't use Text anymore, but each time we use it we would need to style it. I would prefer to have a set of typography components instead, or use those in paper, after customizing paper's fonts in the theme

App.js
richiemccoll: Yep 🤦 . Done https://github.com/nearform/optic-expo/pull/21/commits/6974fcc9fa6692c17967c246fb4aa0e9260ed39a

components/Auth.js
richiemccoll: Good point. I tried this earlier but I was missing the wrapper `Image` component. Here's the commit with the updated UI https://github.com/nearform/optic-expo/pull/21/commits/9477af4da284bbc74add4267e54bc65ec1a97330

components/typography.js
n370: Sounds like something we can disclose at the project README file. Thoughts?

n370 (approved):

n370:
Let's wait for Simone's final review though

simoneb (changes requested):
LGTM apart from the layout. On a smaller screen see how the logo is close to the bottom, or beyond.

image

I would suggest to use a column flex layout (which is the default) with justifyContent: space-around and alignItems: center, with three children equally sized (flex = 1). 

the first contains the title and subtitle, the second contains the button, the third contains the logo.

not sure it will look great (please try) but at least it doesn't suffer from the spacing issues.

don't use spacer basically

components/typography.js
richiemccoll: I think we could definitely do with some documentation (in general).

richiemccoll:
@simoneb I've updated the container to use flex (https://github.com/nearform/optic-expo/pull/21/commits/8d3ff81ac23b3c0c7159d4c64552dd45558402b6). Could you please check again when you have time?

simoneb (approved):
perfect

PR #23, opened 2021-02-18
feat: update blank tokens screen to match design

richiemccoll:
As I was working on the scan a QR code feature, I updated the Blank tokens UI to match the design. I thought it would make sense to separate these two changes out and land them separately.

- Added a couple of new typography styles. `Headline5` and `Caption` typography style (from figma)
- I've updated the `Headline` component to accept a `level` prop. (Let me know if you feel strongly against this)

### DIFF
--- a/components/Actions.js
+++ b/components/Actions.js
@@ -1,13 +1,15 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { View, StyleSheet } from 'react-native'
 import { IconButton } from 'react-native-paper'
 
 import theme from '../lib/defaultTheme'
 
+import IconButtonWithLabel from './IconButtonWithLabel'
+
 const A11Y_LABELS = {
-  scan: 'Add new secret by scanning a QR Code',
-  upload: 'Add new secret by uploading a QR Code',
-  type: 'Add new secret by typing the details',
+  scan: 'Scan QR Code',
+  upload: 'Upload',
+  type: 'Add details manually',
 }
 
 export default function Actions({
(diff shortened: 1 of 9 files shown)


n370 (approved):
Looks good!

components/Actions.js
```diff
@@ -1,59 +1,68 @@
...
+            onPress={onUploadNewSecretScreen}
+            icon="upload"
+          />
+          <IconButtonWithLabel
+            label={A11Y_LABELS.type}
+            onPress={onTypeNewSecretScreen}
+            icon="import"
+          />
+        </View>
+      )}
```
simoneb, line 41: it would be nice to include an animation at some point. if you prefer not to do it here, can you please create an issue to track it?

components/EmptyTokensText.js
```diff
@@ -1,23 +1,24 @@
...
 export default function EmptyTokensText() {
   return (
     <View style={styles.description}>
-      <Subheading>{UI_STRINGS.heading}</Subheading>
+      <Headline level="5" color={theme.colors.grey}>
+        {UI_STRINGS.heading}
+      </Headline>
       <Spacer size={2} />
-      <Subheading>{UI_STRINGS.description}</Subheading>
+      <BodyText color={theme.colors.grey}>{UI_STRINGS.description}</BodyText>
```
simoneb, line 21: because all the text except the one on dark background has this color, shall we not use the textColor from the theme (or whatever it's called). or maybe, even make it the default on our custom typography in the same way as it's done for the paper typography

lib/defaultTheme.js
```diff
@@ -5,6 +5,7 @@ const theme = {
   colors: {
     ...DefaultTheme.colors,
     primary: '#2165E3',
+    grey: '#6D6D68',
```
simoneb, line 8: as mentioned before, I believe this is the text color, already existing in the theme

simoneb (commented):
A couple of comments

components/EmptyTokensText.js
richiemccoll: The `text` color from the default theme in `react-native-paper` points to `#000000`.  https://github.com/callstack/react-native-paper/blob/main/src/styles/DefaultTheme.tsx#L15. I'll make it the default for custom typography.

components/Actions.js
richiemccoll: Yes, good shout. I'll create a ticket to add micro-interactions.

lib/defaultTheme.js
```diff
@@ -5,6 +5,7 @@ const theme = {
   colors: {
     ...DefaultTheme.colors,
     primary: '#2165E3',
+    text: '#6D6D68',
```
simoneb, line 8: ok sorry I have probably misled you a bit on this. the default color in paper is black, but then you'll see that they'll make it lighter in the typography components. doing this here will make our text match this color, but the text elements created using paper typography components lighter than this. I hope this is clear now

simoneb (approved):

lib/defaultTheme.js
richiemccoll: I've updated the typography components to use `color` module internally. Similar to how `react-native-paper` approached it https://github.com/callstack/react-native-paper/blob/main/src/components/Typography/StyledText.tsx#L16. Here's the commit https://github.com/nearform/optic-expo/pull/23/commits/1d392074239c7566db1f22bce50b35cf966167bc

PR #24, opened 2021-02-18
chore: enforce identifier definition order

n370:
A little housekeeping work here based on few comments and slack discussions around #13.

Effectively this PR do:

- Add the "no-use-before-define" eslint rule to our configuration to aid on enforcing the consensus arrived regarding the order in which ui strings and styles constants are defined in component files.
- Update event handler function name in `<TypeNewSecret />` function to a more meaningful one.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,7 +17,8 @@
   },
   "plugins": ["react"],
   "rules": {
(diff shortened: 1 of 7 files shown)


simoneb (approved):
fine for me if you agreed on this

richiemccoll (approved):
LGTM

PR #25, opened 2021-02-18
feat: use react-native storage

### LINKED ISSUE #19: use RN storage for storing secrets

n370:
Closes #19

This PR introduces the following changes:

- Replace in-memory secret storage with the less ephemeral solution provided by the [AsyncStorage](https://docs.expo.io/versions/v40.0.0/sdk/async-storage/) RN library.
- Introduces a security layer around the storage solution using the [SecureStore](https://docs.expo.io/versions/latest/sdk/securestore/) package where available and a symmetrical encryption strategy with the cryptography tools provided by the [aes-js](https://www.npmjs.com/package/aes-js) library.
- Configure a new husky hook to run the project's unit tests before code is pushed upstream.

### DIFF
--- a/components/Main.js
+++ b/components/Main.js
@@ -4,7 +4,10 @@ import { createStackNavigator } from '@react-navigation/stack'
 
 import theme from '../lib/defaultTheme'
 import routes from '../lib/routeDefinitions'
+import { useAuthenticationContext } from '../context/authentication'
+import { useSecrets } from '../context/secrets'
 
+import Auth from './Auth'
 import Home from './screens/Home'
 import TypeNewSecretScreen from './screens/TypeNewSecretScreen'
 import ScanNewSecretScreen from './screens/ScanNewSecretScreen'
@@ -32,6 +35,14 @@ const UI_STRINGS = {
 }
 
 export default function Main() {
+  const { user, handleLogin } = useAuthenticationContext()
+
+  const { isInitialized } = useSecrets()
+
+  if (!user || !isInitialized) {
+    return <Auth {...{ handleLogin }} />
+  }
+
   return (
     <NavigationContainer>
       <Stack.Navigator
--- a/components/Main.js
+++ b/components/Main.js
@@ -4,7 +4,7 @@ import { createStackNavigator } from '@react-navigation/stack'
 
 import theme from '../lib/defaultTheme'
 import routes from '../lib/routeDefinitions'
-import { useAuthenticationContext } from '../context/authentication'
+import { useAuthentication } from '../context/authentication'
 import { useSecrets } from '../context/secrets'
 
 import Auth from './Auth'
@@ -35,7 +35,7 @@ const UI_STRINGS = {
 }
 
 export default function Main() {
-  const { user, handleLogin } = useAuthenticationContext()
+  const { user, handleLogin } = useAuthentication()
 
   const { isInitialized } = useSecrets()
 
--- a/components/Main.js
+++ b/components/Main.js
@@ -1,17 +1,26 @@
 import React from 'react'
 import { NavigationContainer } from '@react-navigation/native'
 import { createStackNavigator } from '@react-navigation/stack'
+import {
+  useFonts as usePoppins,
+  Poppins_700Bold,
+} from '@expo-google-fonts/poppins'
+import {
+  useFonts as useDidactGothic,
+  DidactGothic_400Regular,
+} from '@expo-google-fonts/didact-gothic'
+import AppLoading from 'expo-app-loading'
 
 import theme from '../lib/defaultTheme'
 import routes from '../lib/routeDefinitions'
 import { useAuthentication } from '../context/authentication'
 import { useSecrets } from '../context/secrets'
 
-import Auth from './Auth'
 import Home from './screens/Home'
 import TypeNewSecretScreen from './screens/TypeNewSecretScreen'
 import ScanNewSecretScreen from './screens/ScanNewSecretScreen'
 import UploadNewSecretScreen from './screens/UploadNewSecretScreen'
+import Auth from './Auth'
 import HomeHeaderRight from './HomeHeaderRight'
 import DefaultHeaderLeft from './DefaultHeaderLeft'
 
@@ -35,11 +44,23 @@ const UI_STRINGS = {
 }
 
 export default function Main() {
-  const { user, handleLogin } = useAuthentication()
+  const [hasPoppinsLoaded] = usePoppins({
+    Poppins_700Bold,
+  })
+
+  const [hasDidactLoaded] = useDidactGothic({
+    DidactGothic_400Regular,
+  })
 
   const { isInitialized } = useSecrets()
 
-  if (!user || !isInitialized) {
+  const { user, handleLogin } = useAuthentication()
+
+  if (!hasDidactLoaded || !hasPoppinsLoaded || !isInitialized) {
+    return <AppLoading />
+  }
+
+  if (!user) {
     return <Auth {...{ handleLogin }} />
   }
 
--- a/context/secrets.js
+++ b/context/secrets.js
@@ -7,31 +7,45 @@ import React, {
   useEffect,
 } from 'react'
 
-import { upsert, find } from '../lib/secretsManager'
+import SecretsManager from '../lib/secretsManager'
 
 const SecretsContext = createContext({})
 
-export function useSecretsContext() {
+export function useSecrets() {
   return useContext(SecretsContext)
 }
 
 export function SecretsProvider({ children }) {
+  const [isInitialized, setInitialization] = useState(false)
   const [secrets, setSecrets] = useState([])
 
-  useEffect(() => {
-    ;(async () => {
-      setSecrets(await find())
-    })()
-  }, [])
+  const secretsManager = useMemo(() => new SecretsManager(), [])
 
-  const add = useCallback(async secret => {
-    await upsert(secret)
-    setSecrets(await find({ uid: secret.uid }))
-  }, [])
+  useEffect(() => {
+    async function initialize() {
+      await secretsManager.initialize()
+      setSecrets(await secretsManager.find())
+      setInitialization(true)
+    }
+
+    initialize()
+  }, [secretsManager])
+
+  const add = useCallback(
+    async secret => {
+      await secretsManager.upsert(secret)
+      setSecrets(await secretsManager.find({ uid: secret.uid }))
+    },
+    [secretsManager]
+  )
 
   return (
     <SecretsContext.Provider
-      value={useMemo(() => ({ secrets, add }), [secrets, add])}
+      value={useMemo(() => ({ isInitialized, secrets, add }), [
+        isInitialized,
+        secrets,
+        add,
+      ])}
     >
       {children}
     </SecretsContext.Provider>
--- a/context/secrets.js
+++ b/context/secrets.js
@@ -4,6 +4,7 @@ import React, {
   useMemo,
   useContext,
   useCallback,
+  useEffect,
 } from 'react'
 
 import { upsert, find } from '../lib/secretsManager'
@@ -16,6 +17,12 @@ export function useSecretsContext() {
 
 export function SecretsProvider({ children }) {
   const [secrets, setSecrets] = useState([])
+
+  useEffect(() => {
+    ;(async () => {
+      setSecrets(await find())
+    })()
+  }, [])
 
   const add = useCallback(async secret => {
     await upsert(secret)
--- a/lib/secretsManager.js
+++ b/lib/secretsManager.js
@@ -1,79 +1,151 @@
+import { v4 as uuid } from 'uuid'
+import aesjs from 'aes-js'
+import * as SecureStore from 'expo-secure-store'
 import AsyncStorage from '@react-native-async-storage/async-storage'
-import uuid from 'uuid-random'
 
-let secretStorageKey = 'insecure'
+const encryptionPassphraseStorageKey = 'encryptionPassphrase'
+const secretStorageKey = 'secrets'
 
-/**
- * Find all secrets, applying a selector if provided.
- * When provided, all where clause must be fulfilled (logical AND).
- * By default, returns all secrets
- * @async
- * @param {Object} where - hash of where clauses returned secrets must fulfill with
- * @returns {Array<Object>} list (may be empty) of matching secrets
- */
-export async function find(where = {}) {
-  let secrets = []
-
-  try {
-    const props = Object.keys(where)
-    secrets = JSON.parse(
-      (await AsyncStorage.getItem(secretStorageKey)) || '[]'
-    ).filter(secret => props.every(prop => secret[prop] === where[prop]))
-  } catch (err) {
-    console.error('Failed to read secrets from data store', err)
+export default class SecretsManager {
+  constructor() {
+    this.encryptionPassphrase = null
   }
 
-  return secrets
-}
+  async initialize() {
+    const [
+      getEncryptionPassphrase,
+      setEncryptionPassphrase,
+    ] = (await SecureStore.isAvailableAsync())
+      ? [SecureStore.getItemAsync, SecureStore.setItemAsync]
+      : [AsyncStorage.getItem, AsyncStorage.setItem]
 
-/**
- * Add or update a secret to the list.
- * If not provided, an _id key will be generated
- * @async
- * @param {Object} secret - secret to upsert
- * @returns {Object} the added/updated secret
- */
-export async function upsert(secret) {
-  const secretId = secret._id || uuid()
-  const secrets = await find()
-  let upserted = secrets.find(({ _id }) => _id === secretId)
+    const encryptionPassphrase = await getEncryptionPassphrase(
+      encryptionPassphraseStorageKey
+    )
 
-  if (upserted) {
-    Object.assign(upserted, secret)
-  } else {
-    upserted = { _id: secretId, ...secret }
-    secrets.push(upserted)
+    if (encryptionPassphrase) {
+      this.encryptionPassphrase = JSON.parse(encryptionPassphrase)
+      return
+    }
+
+    this.encryptionPassphrase = this.generateEncryptionPassphrase()
+
+    await setEncryptionPassphrase(
+      encryptionPassphraseStorageKey,
+      JSON.stringify(this.encryptionPassphrase)
+    )
+
+    await this.set([])
   }
 
-  try {
-    await AsyncStorage.setItem(secretStorageKey, JSON.stringify(secrets))
-  } catch (err) {
-    console.error('Failed to add new secret to data store', err)
+  generateEncryptionPassphrase() {
+    return Array.from(crypto.getRandomValues(new Uint8Array(32)))
   }
 
-  return upserted
-}
-
-/**
- * Remove a secret from the list. Does not fail if it can not be found.
- * @async
- * @param {String} secretId - removed secret id
- * @returns {Object} the removed secret (may be null)
- */
-export async function remove(secretId) {
-  const secrets = await find()
-  const idx = secrets.findIndex(({ _id }) => _id === secretId)
-  let removed = null
-
-  if (idx !== -1) {
-    removed = secrets.splice(idx, 1)[0]
+  encrypt(secrets) {
+    const secretsBytes = aesjs.utils.utf8.toBytes(secrets)
+    const aesCtr = new aesjs.ModeOfOperation.ctr(
+      Uint8Array.from(this.encryptionPassphrase),
+      new aesjs.Counter(5)
+    )
+    const encryptedBytes = aesCtr.encrypt(secretsBytes)
+    return aesjs.utils.hex.fromBytes(encryptedBytes)
   }
 
-  try {
-    await AsyncStorage.setItem(secretStorageKey, JSON.stringify(secrets))
-  } catch (err) {
-    console.error('Failed to remove secret from data store', err)
+  decrypt(encryptedSecrets) {
+    const encryptedBytes = aesjs.utils.hex.toBytes(encryptedSecrets)
+    const aesCtr = new aesjs.ModeOfOperation.ctr(
+      Uint8Array.from(this.encryptionPassphrase),
+      new aesjs.Counter(5)
+    )
+    const decryptedBytes = aesCtr.decrypt(encryptedBytes)
+    return aesjs.utils.utf8.fromBytes(decryptedBytes)
   }
 
-  return removed
+  async get() {
+    const encrypted = await AsyncStorage.getItem(secretStorageKey)
+    return JSON.parse(this.decrypt(encrypted))
+  }
+
+  async set(secrets) {
+    return AsyncStorage.setItem(
+      secretStorageKey,
+      this.encrypt(JSON.stringify(secrets))
+    )
+  }
+
+  /**
+   * Find all secrets, applying a selector if provided.
+   * When provided, all where clause must be fulfilled (logical AND).
+   * By default, returns all secrets
+   * @async
+   * @param {Object} where - hash of where clauses returned secrets must fulfill with
+   * @returns {Array<Object>} list (may be empty) of matching secrets
+   */
+  async find(where = {}) {
+    let secrets = []
+
+    try {
+      const whereClauses = Object.keys(where)
+      secrets = (await this.get()).filter(secret =>
+        whereClauses.every(clause => secret[clause] === where[clause])
+      )
+    } catch (err) {
+      console.error('Failed to read secrets from data store', err)
+    }
+
+    return secrets
+  }
+
+  /**
+   * Add or update a secret to the list.
+   * If not provided, an _id key will be generated
+   * @async
+   * @param {Object} secret - secret to upsert
+   * @returns {Object} the added/updated secret
+   */
+  async upsert(secret) {
+    const secretId = secret._id || uuid()
+    const secrets = await this.get()
+    let upserted = secrets.find(({ _id }) => _id === secretId)
+
+    if (upserted) {
+      Object.assign(upserted, secret)
+    } else {
+      upserted = { _id: secretId, ...secret }
+      secrets.push(upserted)
+    }
+
+    try {
+      await this.set(secrets)
+    } catch (err) {
+      console.error('Failed to upsert secret into data store', err)
+    }
+
+    return upserted
+  }
+
+  /**
+   * Remove a secret from the list. Does not fail if it can not be found.
+   * @async
+   * @param {String} secretId - removed secret id
+   * @returns {Object} the removed secret (may be null)
+   */
+  async remove(secretId) {
+    const secrets = await this.get()
+    const idx = secrets.findIndex(({ _id }) => _id === secretId)
+    let removed = null
+
+    if (idx !== -1) {
+      removed = secrets.splice(idx, 1)[0]
+    }
+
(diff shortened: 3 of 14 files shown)


n370:
@simoneb I didn't moved on with the secure implementation yet because I got confused by a few things I'd like to talk with you. I'll ping you on slack.

simoneb:
No problem. If you prefer, you could start out by using the non-secure storage only and then we create an issue to handle the secrets securely. At this stage of development encrypting them is not the first priority.

n370:
Hey @simoneb I think I got a secure secrets storage implementation working. I spent a fair amount of the day trying a bunch of crypto libraries without much success until I found this one https://www.npmjs.com/package/@meeco/cryppo which is working fine except for a noticeable degradation on the UI responsiveness when reading and writing data. 

I believe this is due to the fact that I designed things in such a way that every read and every write operation encrypts/decrypts the secrets collection which probably isn't the most elegant way of doing this. 

Anything different from it though will demand more time to design, test and implement so I'll leave it up to you to decide if we should now finally branch this off or keep trying to improve it. Maybe I just chose a terrible crypto lib and changing it speed things up for the user. Worst case scenario we'll need to think about a more elegant solution that encrypt/decrypt things only when the app is loaded and when it is closed or simply add loading indicators to the UI when read/write actions are performed by the user.

context/secrets.js
```diff
@@ -17,9 +18,17 @@ export function useSecretsContext() {
 export function SecretsProvider({ children }) {
   const [secrets, setSecrets] = useState([])
 
+  useEffect(() => {
+    ;(async () => {
+      const secretsManager = await SecretsManager.getInstance()
```
simoneb, line 23: let's avoid singletons. if we want a single instance of this object let's create it once and reuse it but avoid this getInstance pattern

lib/secretsManager.js
```diff
@@ -1,77 +1,173 @@
...
+  async initializeWithSecureRandomPassphrase() {
+    const encryptionKey = await SecureStore.getItemAsync(
+      encryptionPassphraseStorageKey
     )
-  } catch (err) {
-    console.error('Failed to read secrets from data store', err)
+
+    if (encryptionKey) {
+      this.encryptionKey = encryptionKey
+    } else {
```
simoneb, line 47: minor: you can easily avoid if/else if you return early. makes the code less nested and easier to read

lib/secretsManager.js
```diff
@@ -1,77 +1,173 @@
...
+  constructor() {
+    this.encryptionKey = null
+  }
+
+  async init() {
+    if (SecureStore.isAvailableAsync()) {
+      await this.initializeWithSecureRandomPassphrase()
+    } else {
+      await this.initializeWithVulnerablePassphrase()
+    }
```
simoneb, line 37: I'm not sure if we even need to support this. It's a graceful fallback, but nevertheless we could generate a random key instead of an hardcoded one

simoneb (changes requested):
Overall the approach is sound but we need to make some changes, besides the inline comments:

- I'm not sure about the choice of library. When choosing a library, and especially one which deals with encryption, I'd much rather stick to popular options. I usually evaluate that using the number of downloads in npm and the github stars. Both of these stats are not very exciting for this library. It may be convenient to use, but I would basically not trust it
- We need to change the approach in terms how when we encrypt and decrypt. These are expensive operations. I have no previous experience with these operations on a mobile device so I can't say how expensive they really are, but I guess a more conservative approach would be to "unseal" the secrets when the application starts and use them unsealed, then "flush" them and seal them on demand, e.g. when a new secret is added. By doing this, we're basically not doing any cryptographic operations except at application startup or when we need to write/modify a secret

context/secrets.js
n370: Fair enough! I'll make the adjustments.

lib/secretsManager.js
n370: Yup! I can change that too.

lib/secretsManager.js
n370: Makes sense to me! I'll make the adjustments or remove it all together.

n370:
> Overall the approach is sound but we need to make some changes, besides the inline comments:
> 
>     * I'm not sure about the choice of library. When choosing a library, and especially one which deals with encryption, I'd much rather stick to popular options. I usually evaluate that using the number of downloads in npm and the github stars. Both of these stats are not very exciting for this library. It may be convenient to use, but I would basically not trust it
> 
>     * We need to change the approach in terms how when we encrypt and decrypt. These are expensive operations. I have no previous experience with these operations on a mobile device so I can't say how expensive they really are, but I guess a more conservative approach would be to "unseal" the secrets when the application starts and use them unsealed, then "flush" them and seal them on demand, e.g. when a new secret is added. By doing this, we're basically not doing any cryptographic operations except at application startup or when we need to write/modify a secret

Yeah, I agree with you. Let me see what I can do about it today.

context/secrets.js
```diff
@@ -4,22 +4,33 @@ import React, {
...
   useCallback,
+  useEffect,
 } from 'react'
 
-import { upsert, find } from '../lib/secretsManager'
+import SecretsManager from '../lib/secretsManager'
+
+const secretsManager = new SecretsManager()
+
+;(async () => await secretsManager.init())()
```
simoneb, line 14: let's avoid this pattern please, find a more idiomatic way to do this instead. if we need really async initialization, put it in application startup and display splash screen until the secrets manager is initialized rather than coming up with fancy ways to do it. plus, this probably doesn't do what you think it does, it most likely suffers from a race condition because the code is not actually waiting for the async initialization to complete before moving on. this probably only seems to work because of timing in how things are executed in the app. make this async initialization take longer, and the app will blow up

context/secrets.js
```diff
@@ -4,22 +4,33 @@ import React, {
...
   return useContext(SecretsContext)
 }
 
 export function SecretsProvider({ children }) {
   const [secrets, setSecrets] = useState([])
 
+  useEffect(() => {
+    ;(async () => {
+      setSecrets(await secretsManager.find())
+    })()
```
simoneb, line 28: avoid this pattern. if you need to run something async wrap it in a function and invoke it from within the useeffect hook

lib/secretsManager.js
```diff
@@ -1,77 +1,154 @@
...
+
+    await this.set([])
   }
 
-  return secrets
-}
+  generateEncryptionPassphrase() {
+    return JSON.stringify(
+      Array.from({ length: 32 }).map(() => Math.floor(Math.random() * 256))
+    )
```
simoneb, line 45: this is not a cryptographic secure way to generate a random key. I appreciate this is probably not a big deal in this specific context, but since we're following good cryptographic patterns, let's just follow them completely. you need to find something that's equivalent to the Node.js crypto library.

lib/secretsManager.js
```diff
@@ -1,77 +1,154 @@
...
+  }
+
+  async encrypt(secrets) {
+    const secretsBytes = aesjs.utils.utf8.toBytes(JSON.stringify(secrets))
+    const aesCtr = new aesjs.ModeOfOperation.ctr(
+      Uint8Array.from(JSON.parse(this.encryptionPassphrase)),
+      new aesjs.Counter(5)
```
simoneb, line 52: does this create always the same thing? if so, let's extract it in a module scoped constant and give it a meaningful name

simoneb (changes requested):

context/secrets.js
n370: The useEffect callback can't be an async function so the alternative I guess would be to use the promise syntax. Is that what you mean?

context/secrets.js
n370: You're correct, this pattern doesn't make any sense at this place. It is suited though to the useEffect callback body inside the secrets provider function. See: https://github.com/nearform/optic-expo/pull/25#discussion_r580632889 and the following gist:

```js
  // ...
  
  const secretsManager = useMemo(() => new SecretsManager(), [])

  const [secrets, setSecrets] = useState([])

  useEffect(() => {
    ;(async () => {
      await secretsManager.init()
      setSecrets(await secretsManager.find())
    })()
  }, [secretsManager])

 // ...
```

lib/secretsManager.js
n370: I have to admit I'm puzzled by this task. Here I just kind of followed the docs at https://www.npmjs.com/package/aes-js#keys. I explored a bit with the possibility of using https://www.npmjs.com/package/pbkdf2, https://www.npmjs.com/package/scrypt-js or https://www.npmjs.com/package/bcryptjs but I couldn't quite figure how to put one of them together with aes-js. pbkdf2 for instance appears to rely on the native nodejs crypto lib and I couldn't get it to work on a first try. bcryptjs and scrypt-js both demand some kind of a input to be turned into a hash and then their output must be a char string of 16, 24 or 32 bytes to be accepted by the aes-js aes cipher functions which I couldn't figure out exactly as well on a first try. https://www.npmjs.com/package/aes-js#ctr---counter-recommended is the example I followed to implement this iteration.

lib/secretsManager.js
n370: It seems to be something we really need to create and recreate every time.

lib/secretsManager.js
simoneb: The approach you followed makes sense, the only bit that's missing, which as I said is probably a relatively minor detail in these specific circumstances even though it would be a pity if we missed this step, is that the key you use should be cryptocraphically secure. In Node.js you would use crypto.randomBytes to generate such a key, but using Math.random is not the same thing.

Now, because in AES the keys need to be of a specific length, you have 2 options:

- generate a truly random key of the required length
- use a password-based key derivation function which, given a password string of random length and not necessarily cryptographically secure, gives you a cryptographically secure key (script, bcrypt, pdkdf2)

Clearly, in option 2, you still have the issue of generating that password, but luckily we don't need that, because there's an extremely popular library which does the same as Node's randomBytes for react native: https://www.npmjs.com/package/react-native-get-random-values. This also seems to work fine in expo, so that should do

package.json
```diff
@@ -14,15 +14,19 @@
   "dependencies": {
     "@expo-google-fonts/didact-gothic": "^0.1.0",
     "@expo-google-fonts/poppins": "^0.1.0",
+    "@react-native-async-storage/async-storage": "1.13.4",
     "@react-native-community/masked-view": "0.1.10",
     "@react-navigation/native": "5.9.2",
     "@react-navigation/stack": "5.14.2",
+    "aes-js": "3.1.2",
+    "bcryptjs": "2.4.3",
```
simoneb, line 22: cleanup the packages if we're not using bcrypt

package.json
```diff
@@ -36,7 +40,8 @@
     "react-native-screens": "~2.15.2",
     "react-native-svg": "^12.1.0",
     "react-native-web": "~0.13.12",
-    "url-otpauth": "^2.0.0"
+    "url-otpauth": "^2.0.0",
+    "uuid-random": "1.3.2"
```
simoneb, line 44: prefer the more popular https://www.npmjs.com/package/uuid

context/secrets.js
simoneb: Ok I guess it's my personal preference but I really do not like this iife pattern in useEffect. I would prefer something less exotic, which works the same:

```suggestion
    async function storeSecrets() {
      setSecrets(await secretsManager.find())
    }
    storeSecrets()
```

context/secrets.js
simoneb: aside from the syntax, this approach will still not work. this is not preventing the application from trying to access an uninitialized secrets manager. if the initialization is async and we can't do otherwise, it should be moved to the root of  the component tree, in the entry point of the application, and delay application startup until the initialization is complete, then inject the secretsManager instance via react context. if you put it here, the application may try to operate on an uninitialized instance and there's nothing you can do to prevent that

n370: I see what you mean now. I'll fix it right away.

context/secrets.js
n370: That's fine too! I'll change it!

package.json
n370: Done!

package.json
n370: Done!

lib/secretsManager.js
n370: I updated the code with a new version that uses the `react-native-get-random-values` lib. Thanks for clearing things out a bit for me here.

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
-  const secretId = secret._id || DUMMY_UUID_GEN++
-  const secrets = await find()
-  let upserted = secrets.find(({ _id }) => _id === secretId)
-
-  if (upserted) {
-    Object.assign(upserted, secret)
-  } else {
-    upserted = { _id: secretId, ...secret }
-    secrets.push(upserted)
+  async encrypt(secrets) {
```
simoneb, line 66: remove async

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+    return JSON.stringify(
+      Array.from(
+        crypto.getRandomValues(
+          Uint8Array.from(Array.from({ length: 32, value: 0 }))
+        )
+      )
+    )
+  }
+
+  async flush() {
```
simoneb, line 52: not sure about the naming. flush usually means persisting temporary changes. it looks like you're cleaning stuff up here instead

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
-    console.error('Failed to read secrets from data store', err)
   }
 
-  return secrets
-}
+  generateEncryptionPassphrase() {
+    return JSON.stringify(
+      Array.from(
+        crypto.getRandomValues(
+          Uint8Array.from(Array.from({ length: 32, value: 0 }))
```
simoneb, line 46: you don't need all this ceremony to create a Uint8Array of a certain size

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+  async initialize() {
+    let getEncryptionKey = AsyncStorage.getItem
+    let setEncryptionKey = AsyncStorage.setItem
+
+    if (SecureStore.isAvailableAsync()) {
+      getEncryptionKey = SecureStore.getItemAsync
+      setEncryptionKey = SecureStore.setItemAsync
+    }
+
+    const encryptionPassphrase = await getEncryptionKey(
```
simoneb, line 23: nit: naming consistency. if the function is called getEncryptionKey, why are we assigning it to a variable called encryptionPassphrase? Is it a key or a passphrase?

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+  }
+
+  async initialize() {
+    let getEncryptionKey = AsyncStorage.getItem
+    let setEncryptionKey = AsyncStorage.setItem
+
+    if (SecureStore.isAvailableAsync()) {
+      getEncryptionKey = SecureStore.getItemAsync
+      setEncryptionKey = SecureStore.setItemAsync
+    }
```
simoneb, line 21: nit on syntax, to avoid mutable variables.
major problem in not awaiting the result of isAvailableAsync

```suggestion
    const [getEncryptionKey, setEncryptionKey] = await SecureStore.isAvailableAsync() 
      ? [SecureStore.getItemAsync, SecureStore.setItemAsync] 
      : [AsyncStorage.getItem, AsyncStorage.setItem]
```

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+      Array.from(
+        crypto.getRandomValues(
+          Uint8Array.from(Array.from({ length: 32, value: 0 }))
+        )
+      )
+    )
+  }
+
+  async flush() {
+    if (SecureStore.isAvailableAsync()) {
```
simoneb, line 53: you must await this

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
-
-  if (upserted) {
-    Object.assign(upserted, secret)
-  } else {
-    upserted = { _id: secretId, ...secret }
-    secrets.push(upserted)
+  async encrypt(secrets) {
+    const secretsBytes = aesjs.utils.utf8.toBytes(JSON.stringify(secrets))
+    const aesCtr = new aesjs.ModeOfOperation.ctr(
+      Uint8Array.from(JSON.parse(this.encryptionPassphrase)),
```
simoneb, line 69: why is encryptionPassphrase a json string? I think we should parse it to a JavaScript value when we initialize/read it, so that the rest of the code in this class doesn't have to deal with JSON

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
- */
-export async function remove(secretId) {
-  const secrets = await find()
-  const idx = secrets.findIndex(({ _id }) => _id === secretId)
-  let removed = null
-
-  if (idx !== -1) {
-    removed = secrets.splice(idx, 1)[0]
+    try {
+      const props = Object.keys(where)
```
simoneb, line 107: give this variable a more meaningful name

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+      secrets = (await this.get()).filter(secret =>
+        props.every(prop => secret[prop] === where[prop])
+      )
+    } catch (err) {
+      console.error('Failed to read secrets from data store', err)
```
simoneb, line 112: not for now, but we may want to improve error handling at some point. please create an issue for this

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+  /**
+   * Add or update a secret to the list.
+   * If not provided, an _id key will be generated
+   * @async
+   * @param {Object} secret - secret to upsert
+   * @returns {Object} the added/updated secret
+   */
+  async upsert(secret) {
+    const secretId = secret._id || uuid()
+    const secrets = await this.find()
```
simoneb, line 127: find without params isn't the same as calling get?

lib/secretsManager.js
```diff
@@ -1,77 +1,169 @@
...
+    let upserted = secrets.find(({ _id }) => _id === secretId)
```
simoneb, line 128: since we have a find function which accepts parameters, shall we not use that one? just to double check, are we using find with parameters anywhere?

simoneb (commented):
This is looking better. I'm still not 100% sure about a tiny bit of logic in the initialization, but see inline comments. The comments this time around are more focused on specific issues, the overall approach looks better.

lib/secretsManager.js
n370: 🤦 I assumed the literal meaning of the word. I'll change it. Is `reset` fair enough as an alternative?

lib/secretsManager.js
n370: Good catch! will do!

tests/secretsManager.test.js
```diff
@@ -0,0 +1,87 @@
...
+  jest.spyOn(secretsManager, 'set')
+})
+
+afterEach(() => {
+  jest.resetAllMocks()
+})
+
+describe('SecretsManager class', () => {
+  describe('find instance method', () => {
+    test('should return an empty list if no secrets are stored', async () => {
```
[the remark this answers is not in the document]
n370: Yeah, I probably need to add coverage to a few other methods of this class alone. The test cases I added here so far are meant to test only the secrets collection manipulation and for that matter I prefer to keep them simple and mock the other class methods that deal with other concerns. Does that make sense now? Furthermore `this.get` and `this.set` are both basically just calling `AsyncStore` methods with the output of `this.encrypt` or outputting some value to `this.decrypt` to be returned. I believe `this.get` and `this.set` don't need to be tested. `this.encrypt` and `this.decrypt` may also be exempt. Still, a good enough test for all these methods may be to assert the correct third-party functions are being called with the correct arguments, what do you think? We can actually also do the same for `this.generateEncryptionPassphrase`, `this.initialize` and `this.flush` (to be renamed). Please let me know if all this make sense to you. I personally have strong opinions about this one but I'm glad to know what you think about it and change if necessary.

components/Main.js
```diff
@@ -32,6 +35,14 @@ const UI_STRINGS = {
 }
 
 export default function Main() {
+  const { user, handleLogin } = useAuthentication()
+
+  const { isInitialized } = useSecrets()
+
+  if (!user || !isInitialized) {
+    return <Auth {...{ handleLogin }} />
```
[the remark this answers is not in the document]
n370: Good question Simone! Here my reasoning was: If the user is not authenticated, then they should not be allowed to even initialize the data store. Off course behind the scenes there is nothing preventing one from happening after the other but I decided to settle this way to avoid getting stuck here while many other aspects of the implementation were incomplete.

Do you think it is okay if we initialize the secrets data together with the fonts and before the user is authenticated. I'm assuming I should've asked you this before but now that we're on a more stable state with the overall solution this change should be easy enough to make in a few minutes.

lib/secretsManager.js
n370: Ohh I didn't know it. I'll change it!

lib/secretsManager.js
n370: I'll see what I can do about it. I'm not sure changing it will give us any more readable or maintainable alternative. Thanks for bringing it up anyway.

lib/secretsManager.js
n370: Will do!

lib/secretsManager.js
n370: https://github.com/nearform/optic-expo/issues/36

lib/secretsManager.js
n370: Ohh yeah it is! good catch! Turns out most of this code is copied verbatim from optic web and that version doesn't have a get function. I'll change it here for the sake!

lib/secretsManager.js
n370: I'm not sure, this is again code I just copied verbatim from optic web. I'll give it a look and change if necessary. As per using find with parameters, yes we are. Here is an example: https://github.com/nearform/optic-expo/blob/d83abae38ba52af5baeb0ac74268ce046d1a1c97/context/secrets.js#L37-L37

lib/secretsManager.js
n370: Good catch! To be frank I was very confused about how to name the variables and functions related with the encryption key. While some third-party libraries I tried called the encryption key, passphrase, others like aes-js called it key. Since we already had the variable storing the storage key value using the "key" word, I then decided to call the thing passphrase to avoid having a variable like "encryptionKeyStorageKey". I genuinely still don't know if there is a difference between using key or passphrase but for now I'll settle with passphrase and change the function name here.

lib/secretsManager.js
n370: Had to double check this one! you're super correct here. I really thought I read this function was synchronous but yes, it isn't. https://docs.expo.io/versions/latest/sdk/securestore/#returns

n370: And very nice usage of array destructuring on your suggestion. I'll take it.

lib/secretsManager.js
n370: yup, very true. see: https://github.com/nearform/optic-expo/pull/25#discussion_r582387927

lib/secretsManager.js
n370: I've changed it per your request after all. 👍

lib/secretsManager.js
n370: Using reset for now!

n370:
@simoneb I updated the PR once again now after you last review and left a few unresolved discussions for us to sync about tomorrow. Thank you once again for the taking the time to work on this one with me. I'll make sure to grab another task tomorrow morning meanwhile.

lib/secretsManager.js
simoneb: Ok cool, fine then

components/Main.js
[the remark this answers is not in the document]
simoneb: Yes I think the second approach you describe makes more sense. I wouldn't expect the secrets initialization to overlap with the user authentication. That is, at least unless we want to tie the secret to the user, which would kinda make sense but that you're not handling here anyway and would require a more complex implementation. Let's consider secret initialization in the same way as fonts for the time being, then

tests/secretsManager.test.js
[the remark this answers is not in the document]
simoneb: okay fair enough. we can stick to this

lib/secretsManager.js
simoneb: then we're good

components/Main.js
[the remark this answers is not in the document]
n370: Done!

lib/secretsManager.js
```diff
@@ -1,77 +1,161 @@
...
+    )
+
+    if (encryptionPassphrase) {
+      this.encryptionPassphrase = JSON.parse(encryptionPassphrase)
+      return
+    }
+
+    this.encryptionPassphrase = this.generateEncryptionPassphrase()
+
+    await this.reset()
```
simoneb, line 33: do we really need this if we're going to overwrite whatever was there two lines below?

package.json
```diff
@@ -14,29 +14,34 @@
   "dependencies": {
     "@expo-google-fonts/didact-gothic": "^0.1.0",
     "@expo-google-fonts/poppins": "^0.1.0",
+    "@react-native-async-storage/async-storage": "1.13.4",
```
simoneb, line 17: I have not checked but please make sure you installed this package using expo add because I believe it's one of those packages that expo expects to be with a specific version to guarantee support

package.json
```diff
@@ -59,7 +64,8 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "pre-push": "yarn test"
```
simoneb, line 68: I prefer to not have this hook because it slows down the development process. Not a general rule, but for frontend tests, which tend to be slower than backend tests, we can do without

components/Main.js
```diff
@@ -32,6 +44,26 @@ const UI_STRINGS = {
 }
 
 export default function Main() {
+  const [hasPoppinsLoaded] = usePoppins({
+    Poppins_700Bold,
+  })
+
+  const [hasDidactLoaded] = useDidactGothic({
+    DidactGothic_400Regular,
+  })
```
simoneb, line 53: we're not supposed to be loading fonts twice. this is already happening in App.js and should not happen again. App initialization should live in a single place at the root of the component tree

simoneb (changes requested):
Overall looks good but we still need to nail the initialization here. The blocking comment is in Main.js

components/Main.js
n370: Good catch! The font loading on App.js was a left over. I had to move everything out to Main.js to be able to couple them with the secret storage initialization. I'm cleaning up that up now.

package.json
n370: Yeah, fair enough. I'm dropping it.

package.json
n370: I'll double check this!

lib/secretsManager.js
n370: Yeah, this function really started as a solution for me to cleanup the mess while getting the encryption right. I guess we can do good enough by just removing or overwriting the secrets collection here. Thoughts?

simoneb: I agree

simoneb (approved):
LGTM

PR #28, opened 2021-02-19
feat: add secret via QR code scan

richiemccoll:
Closes https://github.com/nearform/optic-expo/issues/4.

This PR adds the functionality to scan in a secret from a OTP QR Code. 
- This also adds some new dependencies:
  - `url-otpauth` used to parse a OTP URL. This gives us `issuer`, `account`, and `secret`
  - `expo-barcode-scanner` does all of the heavy lifting. I've also added a mock version of this to simulate native QR scan events.

https://user-images.githubusercontent.com/12698531/108539313-bd09c000-72d7-11eb-8dd3-10215ba19a74.MP4

### DIFF
--- a/components/IconButtonWithLabel.js
+++ b/components/IconButtonWithLabel.js
@@ -27,7 +27,7 @@ export default function IconButtonWithLabel({ onPress, label, icon }) {
   return (
     <View style={styles.container}>
       <View style={styles.label}>
-        <Caption color={theme.colors.surface}>{label}</Caption>
+        <Caption style={{ color: theme.colors.surface }}>{label}</Caption>
       </View>
 
       <IconButton
--- a/components/IconButtonWithLabel.js
+++ b/components/IconButtonWithLabel.js
@@ -1,0 +1,50 @@
+import React from 'react'
+import { View, StyleSheet } from 'react-native'
+import { IconButton } from 'react-native-paper'
+
+import theme from '../lib/defaultTheme'
+
+import { Caption } from './typography'
+
+export default function IconButtonWithLabel({ onPress, label, icon }) {
+  return (
+    <View style={styles.container}>
+      <View style={styles.label}>
+        <Caption
+          style={{
+            color: theme.colors.surface,
+          }}
+        >
+          {label}
+        </Caption>
+      </View>
+
+      <IconButton
+        icon={icon}
+        accessibilityLabel={label}
+        mode="contained"
+        style={styles.secondaryButton}
+        size={30}
+        color={theme.colors.primary}
+        onPress={onPress}
+      />
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+  label: {
+    backgroundColor: theme.colors.onSurface,
+    borderRadius: 4,
+    padding: theme.spacing(0.75),
+    opacity: 0.6,
+  },
+  secondaryButton: {
+    borderRadius: 100,
+    backgroundColor: theme.colors.surface,
+  },
+})
--- a/components/IconButtonWithLabel.js
+++ b/components/IconButtonWithLabel.js
@@ -6,6 +6,23 @@ import theme from '../lib/defaultTheme'
 
 import { Caption } from './typography'
 
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+  label: {
+    backgroundColor: theme.colors.onSurface,
+    borderRadius: 4,
+    padding: theme.spacing(0.75),
+    opacity: 0.6,
+  },
+  secondaryButton: {
+    borderRadius: 100,
+    backgroundColor: theme.colors.surface,
+  },
+})
+
 export default function IconButtonWithLabel({ onPress, label, icon }) {
   return (
     <View style={styles.container}>
@@ -25,20 +42,3 @@ export default function IconButtonWithLabel({ onPress, label, icon }) {
     </View>
   )
 }
-
-const styles = StyleSheet.create({
-  container: {
-    flexDirection: 'row',
-    alignItems: 'center',
-  },
-  label: {
-    backgroundColor: theme.colors.onSurface,
-    borderRadius: 4,
-    padding: theme.spacing(0.75),
-    opacity: 0.6,
-  },
-  secondaryButton: {
-    borderRadius: 100,
-    backgroundColor: theme.colors.surface,
-  },
-})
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -34,7 +35,8 @@
     "react-native-safe-area-context": "3.1.9",
     "react-native-screens": "~2.15.2",
     "react-native-svg": "^12.1.0",
-    "react-native-web": "~0.13.12"
+    "react-native-web": "~0.13.12",
+    "url-otpauth": "^2.0.0"
```
simoneb, line 39: is this what we're using in Web? I remember we are using a library but not which one

richiemccoll: Yeah, I took this lib and the `parse` utility from here. https://github.com/nearform/optic/blob/master/src/lib/qr-parser.js#L19

components/IconButtonWithLabel.js
```diff
@@ -27,7 +27,7 @@ export default function IconButtonWithLabel({ onPress, label, icon }) {
   return (
     <View style={styles.container}>
       <View style={styles.label}>
-        <Caption color={theme.colors.surface}>{label}</Caption>
+        <Caption style={{ color: theme.colors.surface }}>{label}</Caption>
```
simoneb, line 30: why this change? I would prefer no inline styles and exposing a color property from our typography components which allows overriding whatever default value

components/Main.js
```diff
@@ -56,7 +57,7 @@ export default function Main() {
         />
         <Stack.Screen
           name={routes.scan.name}
-          component={() => <Text>{routes.scan.name} placeholder</Text>}
+          component={ScanNewSecret}
```
simoneb, line 60: can we use the Screen suffix for all components which are screens?

components/ScanNewSecret.js
```diff
@@ -0,0 +1,117 @@
...
+    default: {
+      return state
+    }
+  }
+}
+
+export default function ScanNewSecret() {
+  const [{ scan, permission }, dispatch] = useReducer(scanReducer, initialState)
+  const { user } = useAuthenticationContext()
+  const { add } = useSecretsContext()
```
simoneb, line 65: nit: these hooks could be named without the Context suffix. We don't really care that these give you access to a React context, we only care that we get what we need out of them, which are auth and secrets utilities, respectively

components/Main.js
richiemccoll: Yes, I'll create a `screens` directory inside components as well and move these components into there.

components/IconButtonWithLabel.js
richiemccoll: Fixed https://github.com/nearform/optic-expo/pull/28/commits/595446b3e7eb98d72fa6c3cd102872177de942cb.

components/screens/ScanNewSecretScreen.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export default function ScanNewSecretScreen() {
+  const [{ scan, permission }, dispatch] = useReducer(scanReducer, initialState)
+  const { user } = useAuthenticationContext()
+  const { add } = useSecretsContext()
+  const { navigate } = useNavigation()
+
+  async function requestPermissions() {
+    const { status } = await BarCodeScanner.requestPermissionsAsync()
```
simoneb, line 68: can this throw? if so, is it ok not to handle it?

components/typography.js
```diff
@@ -40,17 +47,29 @@ export function Headline({ children, alpha, style, level, ...props }) {
   )
 }
 
-export function BodyText({ children, alpha, style, ...props }) {
-  const textColor = color(theme.colors.text).alpha(alpha).rgb().string()
+export function BodyText({
+  children,
+  color = theme.colors.text,
+  alpha = 1,
```
simoneb, line 53: minor: the idiomatic react way to set default prop values is via Component.defaultProps

simoneb (commented):
I see 2 issues:

- on my android phone I can't scan a QR code. I can enable the camera but can't go past that, the camera doesn't seem to "see" the QR code. I tried tapping on the screen to see if it required some sort of interaction, but nothing happens
- minor: the text which is displayed when requesting permissions for the camera should either be hidden (do we need it?) or displayed more nicely (centered perhaps?)

simoneb:
this is what I see and can't go past

image

components/typography.js
richiemccoll: I'm happy to change it. However, worth noting that the `defaultProps` API is likely going to be deprecated in future versions. https://github.com/reactjs/rfcs/pull/107.

richiemccoll:
@simoneb I've managed to get this feature working on both iPhone and Android. I've pushed some [changes](https://github.com/nearform/optic-expo/pull/28/commits/c9318f251f7457ef0f9bfbb7c5f050b97362c076) around `requestPermissionsAsync`. This was causing the feature to not boot up on Android. Please check again on Monday to see if it fixes your issue.

components/screens/ScanNewSecretScreen.js
```diff
@@ -0,0 +1,126 @@
...
+import { useAuthenticationContext } from '../../context/authentication'
+import { useSecretsContext } from '../../context/secrets'
+import { parse } from '../../lib/qrParser'
+import routes from '../../lib/routeDefinitions'
+import { BodyText } from '../typography'
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    flexDirection: 'column',
```
simoneb, line 15: this is the default

simoneb (approved):
Looking good now. Will still need some UI tweaks at some point:

- when the camera is opened the app seems to freeze for a small amount of time in a state that has the camera screen semi transparent on top of the home screen
- when scanning the token the speed dial menu remains open

For the current stage of development we don't have to fix those now though. Good job

richiemccoll:
@simoneb I've added a couple of commits since your last review.
- A mock for the `expo-barcode-scanner` module (https://github.com/nearform/optic-expo/pull/28/commits/85577ed67665626d8d05a2128d98e7507ab346d4)
- Closing the speed-dial menu on press (https://github.com/nearform/optic-expo/pull/28/commits/12a60a13cbd5a5cb96f1084c0edcffe3c3147128)

simoneb (approved):
LGTM

tests/Main.test.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
+import 'react-native-gesture-handler/jestSetup'
```
simoneb, line 2: shall we move this to a global test setup file, similarly to what's described [here](https://reactnavigation.org/docs/testing/) (first google result but looks like a good approach)?

richiemccoll: Yeah, I was planning on creating a `jest.setup.js`. I hadn't done it until this point because I was working within the one test file. But I'll add it now.

n370 (approved):
Nice!

Issue #17, opened 2021-02-17
Stale UI state on second login


richiemccoll:
**UPDATE**: the PR which fixes this issue has been merged but it needs an expo SDK release to be available. See https://github.com/expo/expo/pull/11503

I noticed some stale UI state after logging in with Google. This only happens after I log in and logout once. I was able to reproduce it on both iPhone and Android.

https://user-images.githubusercontent.com/12698531/108235942-471e2100-713e-11eb-84d4-d1309478cc81.MP4

## Steps to Reproduce
- Login with Google & logout
- Login again with the same account

## Expected
- User arrives on the Home screen

## Actual
- User is stuck on the Auth screen

simoneb:
it works for me

richiemccoll:
@simoneb I've updated the PR description with a screen recording to illustrate the bug.

simoneb:
I tried the steps but could not reproduce it. Let's sync on this tomorrow.

n370:
Seem like we have company here. I've tried the changes seen at https://github.com/expo/expo/issues/11714#issuecomment-765688023 and was able to get a consistent behavior where the login would work every other way.

There is a change coming that may get this one fixed for us: https://github.com/expo/expo/pull/11503#pullrequestreview-599869903.

simoneb:
Good catch!

tiagoheliob:
Issued was fixed on this PR https://github.com/expo/expo/pull/11503

simoneb:
@tiagoheliob the issue was fixed in expo but did you check if we're using the expo version which includes that fix? Because I don't think we updated expo since a long time, hence the fix is not in the version we're using here. I'm reopening the issue, if you want to close it can you please double check this?

simoneb:
I've also created https://github.com/nearform/optic-expo/issues/153 to upgrade expo to latest

simoneb:
We need to verify whether this is still an issue after the application underwent further changes and after the expo upgrade, which may have got us the fix.

salmanm:
Looks like this is fixed now. I tried logging in/out and all seemed fine on my phone

PR #41, opened 2021-03-01
Fix | Login flow fails if performed immediately after logging out

n370:
Closes: #17 

This PR adds:

- [x] Wraps third-party async functions in try/catch blocks where appropriate setting the pattern for proper error handling in the future.
- [x] Preload static assets using the `expo-asset` library.
- [ ] Bumps `expo-auth-session` library to new version that fixes the login issue.

### DIFF
--- a/components/Auth.js
+++ b/components/Auth.js
@@ -4,6 +4,7 @@ import { Platform, View, StyleSheet, Image } from 'react-native'
 import { Button } from 'react-native-paper'
 
 import theme from '../lib/defaultTheme'
+import { useAuthentication } from '../context/authentication'
 
 import { Headline, BodyText } from './typography'
 import Spacer from './Spacer'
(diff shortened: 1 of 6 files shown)


n370:
@simoneb you may want to keep this branch around and use it to add the commit upgrading the auth library when the fixed version is ready. If you do so you can take advantage of a few small improvements I made here while looking for what was causing this issue.

n370:
@simoneb I'm not sure you want this draft to stay around any longer. Did y'all found a solution for this already?

simoneb:
> @simoneb I'm not sure you want this draft to stay around any longer. Did y'all found a solution for this already?

Let's keep it here for reference until it's solved. We know what the issue is but your work here may be useful nonetheless. I'll keep track of this, no worries.

Issue #5, opened 2021-02-14
generate token for secret


simoneb:
the token also needs to be sent to the server

PR #42, opened 2021-03-01
Feat | generate token for secret

n370:
Closes: #5 

This PR adds:

- A back-end service wrapper mostly ported over from optic-web.
- A bare bones `<Secret  />` component with a button to trigger the token generation.

### DIFF
--- a/components/Secret.js
+++ b/components/Secret.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { StyleSheet, View, Text } from 'react-native'
+import { Button } from 'react-native-paper'
+
+import theme from '../lib/defaultTheme'
+
+const styles = StyleSheet.create({
+  container: {
+    margin: theme.spacing(2),
+    borderColor: 'black',
+  },
+})
+
+const UI_STRINGS = {
+  generateTokenButtonLabel: 'Generate Token',
+}
+
+export default function Secret({ data, onGenerateToken }) {
+  const handleGenerateTokenButtonPress = () => onGenerateToken(data)
+
+  return (
+    <View style={styles.container}>
+      <Text key={data._id}>{JSON.stringify(data)}</Text>
+      <Button mode="contained" onPress={handleGenerateTokenButtonPress}>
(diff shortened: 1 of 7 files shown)


components/Secret.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { StyleSheet, View, Text, Button } from 'react-native'
+
+import SecretsService from '../lib/secretsService'
+import { useAuthentication } from '../context/authentication'
+import { useSecrets } from '../context/secrets'
+
+const styles = StyleSheet.create({
+  container: {
+    border: '1px solid black',
```
simoneb, line 10: this is CSS, doesn't work with RN stylesheets

lib/secretsService.js
```diff
@@ -0,0 +1,47 @@
+const apiUrl = '/api' // FIXME
+
+export default class SecretsService {
```
simoneb, line 3: we should change the design of this to avoid having to provide the user. user is in context, we can use context to make sure the service we use to access the API has it without providing it explicitly

components/Secret.js
n370: Yup! this is all gone now!

lib/secretsService.js
n370: Agreed! I'm going with a different design moving forward.

n370:
@simoneb I guess I'm leaving this here for now. The code is functional and does the job of talking to the server to get a token. It works fine. There are a couple of things missing though.

https://github.com/nearform/optic-expo/blob/f06fff2adefcad9ba7eb6f26b177c41990458075/tests/Main.test.js#L107-L155

This unit test is for instance passing but not without logging an error to the console. I spent some time with it trying to figure out what was going on but at this point I think we should consider breaking it down into smaller tests. This one is already doing too much imho.

We're also missing writing new unit tests at least for the new class being added here.

salmanm:
https://github.com/nearform/optic-expo/pull/80 included commits from this PR, so closing...

PR #80, opened 2021-05-04
feat: improve ui and continue from #42

salmanm:
Continuing from https://github.com/nearform/optic-expo/pull/42
Closes https://github.com/nearform/optic-expo/issues/5

- Updated home page UI
- Added features, generate/revoke token
- Added delete secret option

Will add/update some tests later.

https://user-images.githubusercontent.com/2510597/117008305-34ac9000-ad08-11eb-90d4-c162302a32db.mp4

### DIFF
--- a/lib/secretsService.js
+++ b/lib/secretsService.js
@@ -1,0 +1,49 @@
+const apiUrl = 'https://optic-web.herokuapp.com/api'
+
+export default class SecretsService {
+  constructor(user) {
+    this.userIdToken = user.idToken
+  }
+
+  async generateToken(secret) {
+    const response = await fetch(`${apiUrl}/token/${secret._id}`, {
+      method: 'PUT',
+      headers: {
+        authorization: `Bearer ${this.userIdToken}`,
+      },
+    })
+
+    const { token } = await response.json()
+
+    return token
+  }
+
+  async revokeToken(secret) {
+    await fetch(`${apiUrl}/token/${secret._id}`, {
+      method: 'DELETE',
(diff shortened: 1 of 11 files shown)


lib/secretsService.js
```diff
@@ -0,0 +1,49 @@
+const apiUrl = 'https://optic-web.herokuapp.com/api'
```
simoneb, line 1: we may want to make this an env var variable which is embedded in the compiled app instead of hardcoding it

package.json
```diff
@@ -28,7 +28,7 @@
     "expo-secure-store": "~9.3.0",
     "expo-status-bar": "~1.0.3",
     "expo-web-browser": "~8.6.0",
-    "firebase": "^8.4.1",
+    "firebase": "^7.9.0",
```
simoneb, line 31: why the downgrade? because of the expo warning? I've had issues with this in the past whereby an older version of the firebase package wouldn't work, so I accepted the expo warning for the sake of having an up-to-date firebase version

simoneb (approved):
A couple of comments but LGTM overall

lib/secretsService.js
salmanm: Sure, this is the stuff I pulled from the previous PR, will refactor in the next one.

package.json
salmanm: Yes because of warning. And at that time my goal was to get it up and running without warnings/issues to do some development. Now I'll focus on upgrading the SDKs.

Issue #81, opened 2021-05-04
update expo SDK


simoneb:
The app was created and then left there for some time. Let's make sure we're using the latest expo SDK. Upgrade instructions are in the docs. I expect this will be a painless upgrade as everything is relatively recent.

PR #82, opened 2021-05-05
chore: upgrade expo and restructure app

salmanm:
Fixes https://github.com/nearform/optic-expo/issues/81


package.json
```diff
@@ -14,41 +14,42 @@
   "dependencies": {
     "@expo-google-fonts/didact-gothic": "^0.1.0",
     "@expo-google-fonts/poppins": "^0.1.0",
-    "@react-native-async-storage/async-storage": "1.15.1",
+    "@react-native-async-storage/async-storage": "^1.13.0",
```
salmanm, line 17: expo cli did this downgrade

package.json
```diff
@@ -14,41 +14,42 @@
...
-    "expo-status-bar": "~1.0.3",
-    "expo-web-browser": "~8.6.0",
-    "firebase": "^7.9.0",
+    "expo-auth-session": "~3.2.3",
+    "expo-barcode-scanner": "~10.1.2",
+    "expo-random": "~11.1.2",
+    "expo-secure-store": "~10.1.0",
+    "expo-status-bar": "~1.0.4",
+    "expo-web-browser": "~9.1.0",
+    "firebase": "8.2.3",
```
salmanm, line 31: We had 8.4.1 (before I made it 7.9.0) and expo upgrade util automatically made it 8.2.3 so I kept it to avoid firebase warning.

package.json
```diff
@@ -14,41 +14,42 @@
...
+    "react-native-safe-area-context": "3.2.0",
+    "react-native-screens": "~3.0.0",
+    "react-native-svg": "12.1.0",
     "react-native-web": "~0.13.12",
     "url-otpauth": "^2.0.0",
     "uuid": "8.3.2"
   },
   "devDependencies": {
-    "@babel/core": "~7.13.15",
+    "@babel/core": "~7.9.0",
```
salmanm, line 47: Same, expo upgrade util downgraded it. We could up it back, but may be it downgraded for some reason!?

simoneb (commented):
The package versions are fine, in general it's good to stick to whatever expo wants. firebase was an exception really because at some point the version that expo wanted didn't work.

I'm not sure about the whole app restructuring, unless it comes from expo itself. If not, unless you have a good reason to change the app structure I would consider keeping it the same for the sake of remaining consistent with the expo template, possibly making further upgrades easier because there may be some assumptions that expo makes about the file system layout or contents of certain files.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "main": "node_modules/expo/AppEntry.js",
+  "main": "./src/index.js",
```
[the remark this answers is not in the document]
salmanm: No, I wanted to keep all the app related files under `/src` and other stuff outside. So had to do this.

salmanm:
> some assumptions that expo makes about the file system layout or contents of certain files

Yes, and when deviating from that, their docs say to use `registerRootComponent`, which I've done.

> unless you have a good reason to change the app structure

Well, I'm new so no strong opinions here. But to me it makes sense to keep the bootstrapping files separate from the app code. So when upgrading expo, if the root directory has more files in it, they don't clutter the app code structutre.

I mean, we had `.vscode`, `.expo`, etc directories at the same level as `context`, which is very app specific directory. IMHO, it's confusing.

simoneb (approved):

PR #89, opened 2021-05-05
feat: update card ui, add mocked otp display

salmanm:
- [x] Stop login screen from flashing when the auth is checking in the background (show only after it's known that user isn't logged in)
- [x] Clear user from local state after logout
- [x] Removed custom FAB implementation in favor of FAB from `react-native-paper`
- [x] Updated UI to match Figma (might still need another pass)
- [x] Restructured
- [x] Updating tests


simoneb:
Note that there is a known issue around login, just so you know in case you come across it. https://github.com/nearform/optic-expo/issues/17

salmanm:
Yep, also the https://github.com/nearform/optic-expo/pull/41. I've come across this issue and will fix it later.

simoneb (approved):

PR #152, opened 2021-09-28
feat: implementing copy to clipboard

tiagoheliob:
* Implementing re-usable copyable information component
* Removing Edit option from the menu dropdown
* Implementing "Reject" API call

https://user-images.githubusercontent.com/27079992/135158242-81585dae-159b-42b9-a366-43b3f1d5a72c.mp4


src/components/Secret/CopyableInfo.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  iconButton: {
+    padding: 0,
+    marginTop: -10,
+  },
+})
+
+const CopyableInfo = ({ children, textCustomStyle }) => {
+  return (
+    <>
```
simoneb, line 20: you don't need the fragment

simoneb (approved):
Please make sure we only have the yarn.lock file in the repo, not the package-lock

PR #154, opened 2021-09-29
chore(deps): bump expo sdk from 41 to 42 (#153)

tiagoheliob:
* Upgrading expo packages
* Fixing toast implementation after update


src/App.js
```diff
@@ -11,12 +12,14 @@ import Main from './Main'
 export default function App() {
   return (
     <AuthenticationProvider>
-      <SecretsProvider>
-        <PaperProvider theme={theme}>
-          <Main />
-          <StatusBar style="auto" />
-        </PaperProvider>
-      </SecretsProvider>
+      <RootSiblingParent>
```
simoneb, line 15: can you briefly explain this change or point to the docs which explain why it's necessary?

simoneb (approved):
LGTM, see 1 inline comment

src/App.js
tiagoheliob: Yes, when updating the expo it also has updated the clipboard package,  you can check this expo section https://docs.expo.dev/ui-programming/react-native-toast/#cross-platform--react-native-root-toast

simoneb: Makes sense, let's leave this comment here for reference. Looks good, thanks.

PR #167, opened 2021-09-30
chore(deps): ignoring all expo related udpates (83)

tiagoheliob:
* Ignoring all expo related packages since they need to be updated manually

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: 'daily'
     ignore:
-      - dependency-name: '@expo-*'
+      - dependency-name: 'expo-*'
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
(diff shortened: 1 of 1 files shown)


simoneb (changes requested):
This is not accurate enough. If you look at your PR where you upgraded expo, you will see that most packages which expo upgraded have a different name pattern. https://github.com/nearform/optic-expo/pull/154/files

tiagoheliob:
> This is not accurate enough. If you look at your PR where you upgraded expo, you will see that most packages which expo upgraded have a different name pattern. https://github.com/nearform/optic-expo/pull/154/files

@simoneb I might be confused here, but I think they do since most of them are `expo-something-else..`

.github/dependabot.yml
```diff
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: 'daily'
     ignore:
-      - dependency-name: '@expo-'
+      - dependency-name: '@expo-*'
```
simoneb, line 8: this is not the issue. you see the @ at the beginning of the string? :)

simoneb (approved):

PR #184, opened 2021-10-11
fix expo app crash

Eomm:
We suppose the app is crashing due the missing env file

I was not able to collect logs from my android device atm


simoneb (approved):
Let's hope this works :)

simoneb:
@Eomm note when you do commits from the GitHub UI they don't respect commitlint automatically

PR #214, opened 2021-11-05
Bugfix: User secrets

### LINKED ISSUE #213: Bug: User can see secrets belonging to another user
When you switch accounts, you will still see the same secrets. we need to rewrite the logic behind that so it works like in the web app and each user is only able to see secrets that are associated with him/her.

sameer-coder:
This PR fixes issue #213 

Closes #213

### DIFF
--- a/src/context/SecretsContext.tsx
+++ b/src/context/SecretsContext.tsx
@@ -10,12 +10,13 @@ import React, {
 import secretsManager from '../lib/secretsManager'
 import { Secret } from '../types'
 
+import { useAuth } from './AuthContext'
+
 export type ContextType = {
   secrets: Secret[]
   add: (_: Omit<Secret, '_id'>) => Promise<void>
   update: (_: Secret) => Promise<void>
   remove: (_: Secret) => Promise<void>
-  updateUserId: (_: string) => void
 }
 
 const initialContext: ContextType = {
@@ -27,9 +28,6 @@ const initialContext: ContextType = {
     // @todo
   },
   remove: async () => {
-    // @todo
-  },
-  updateUserId: async () => {
     // @todo
   },
 }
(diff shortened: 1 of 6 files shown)


src/lib/secretsManager.ts
```diff
@@ -64,4 +69,4 @@ const remove = async (secretId: string) => {
   }
 }
 
-export default { getAll, get, upsert, remove }
+export default { getAll, getAllByUser, get, upsert, remove }
```
simoneb, line 72: is getAll still needed?

src/screens/HomeScreen.tsx
```diff
@@ -67,7 +67,7 @@ type HomeScreenProps = {
 
 export const HomeScreen: React.FC<HomeScreenProps> = ({ navigation }) => {
   const { user } = useAuth()
-  const { secrets, update, remove } = useSecrets()
+  const { secrets, update, remove, updateUserId } = useSecrets()
```
simoneb, line 70: This is not a great pattern to use. rather, let's pass the user as an argument to this hook, or alternatively make the useSecrets hook use useAuth. 
If there is no user, the operations are invalid and should throw errors or simply do nothing.
If there is a user everything proceeds as normal

simoneb (changes requested):

src/lib/secretsManager.ts
sameer-coder: yes, when saving a new secret we get secrets for all users and push the new one before saving all of them in storage.

src/screens/HomeScreen.tsx
sameer-coder: Yes, I have changed the code so that useAuth is directly used in Secrets. Please review to see if it looks ok now.

src/lib/secretsManager.ts
simoneb: but does it need to be exported?

src/context/SecretsContext.tsx
```diff
@@ -40,29 +42,45 @@ export const SecretsProvider: React.FC<SecretsProviderProps> = ({
   children,
 }) => {
   const [secrets, setSecrets] = useState<Secret[]>([])
+  const { user } = useAuth()
 
   useEffect(() => {
+    let mounted = true
```
simoneb, line 48: what's this?

src/lib/secretsManager.ts
sameer-coder: Right. We don't need to export it

src/context/SecretsContext.tsx
sameer-coder: I was getting an error "Can’t perform a React state update on an unmounted component." which I guess was happening because of state changes while the component was not yet mounted.
I have applied the fix as mentioned [here](https://www.debuggr.io/react-update-unmounted-component/)

simoneb: this will not fix that issue. can you see why?

sameer-coder: Not sure why it won't work.
My understanding was that we are updating the state if the `mounted` variable is `true`, and if the component is un-mounted the `mounted` variable will be set to false and it won't enter the `if` block.

src/context/SecretsContext.tsx
```diff
@@ -40,29 +42,47 @@ export const SecretsProvider: React.FC<SecretsProviderProps> = ({
...
     async function initialize() {
-      setSecrets(await secretsManager.getAll())
+      if (user) {
+        const secrets = await secretsManager.getAllByUser(user.uid)
+
+        if (mounted) setSecrets(secrets)
```
simoneb, line 53:
```suggestion
        if (mounted) {
          setSecrets(secrets)
        }
```

simoneb (approved):

PR #216, opened 2021-11-12
feat: add micro interactions

allforabit:
This PR adds a simple animation to the secret show hide action. 

The other mentioned animations in the issue (route transitions and fab button animations) seem to be already  implemented as can be seen in the screen recording.

https://user-images.githubusercontent.com/537189/141766144-d741ab8a-dad6-410a-9677-a7314ec75aff.mov

- Due to the way the animation works, the secret view is not removed from the component hierarchy and is instead hidden and set to height 0. For this reason, the homescreen snapshots were updated.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,6 +2,7 @@ module.exports = function (api) {
   api.cache(true)
   return {
     presets: ['babel-preset-expo', '@babel/typescript'],
+    plugins: ['react-native-reanimated/plugin'],
     env: {
       production: {
         plugins: ['react-native-paper/babel'],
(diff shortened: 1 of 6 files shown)


simoneb (approved):
Very nice! I've never done animations in RN and it kinda puzzles me that doing such a simple animation requires so much code (at least when compared to CSS transitions), but the effect is really neat 🚀

allforabit:
> Very nice! I've never done animations in RN and it kinda puzzles me that doing such a simple animation requires so much code (at least when compared to CSS transitions), but the effect is really neat 🚀

Thanks! 

It's actually going to get much easier in reanimated 2.30 with the enter and exit transition property. https://docs.swmansion.com/react-native-reanimated/docs/next/api/LayoutAnimations/entryAnimations/ Will be one property.

I think this lib would really come into it's own if you wanted to do really fancy animations with shared elements and the likes.

PR #225, opened 2021-11-16
chore: revert react native paper version

allforabit:
This PR reverts react-native-paper to fix some typing issues (callstack/react-native-paper#2965). In addition it adds typechecking to CI.


package.json
```diff
@@ -69,7 +69,7 @@
     "@types/jest": "^27.0.2",
     "@types/node": "^16.11.6",
     "@types/react": "^17.0.35",
-    "@types/react-native": "0.63.35",
+    "@types/react-native": "0.65.5",
```
simoneb, line 72: why this change?

simoneb (approved):
One comment about a change that doesn't seem to belong here

package.json
allforabit: This is needed to get the typing working, presumably because it's the suitable version for that version of react-native-paper

simoneb: Understood. Let's just consider that we need to be careful about upgrading packages related to expo and react native, because a certain version of expo pins specific versions of those packages. in fact they recommend to use the expo cli when installing packages instead of npm, so it checks the compatibility of your expo version with those packages you're trying to install

allforabit: Ok yep will be very careful about that.

PR #246, opened 2021-11-24
chore: setup expo preview

allforabit:


allforabit:
@simoneb there seems to be some issues with the expo action and the QR code is incorrect. See issue here: https://github.com/expo/expo-preview-action/issues/6 I can find the project link in the command line output and that points here: https://expo.dev/@nearform/optic-expo?release-channel=pr-246 However that doesn't seem to be working on my device.

simoneb:
Before we troubleshoot this can you please align it to how it's done in https://github.com/nearform/optic-expo/blob/master/.github/workflows/publish.yml? E.g. the environment handling

.github/workflows/ci.yml
```diff
@@ -24,3 +24,11 @@ jobs:
...
       - run: yarn typecheck
       - run: yarn test
+
+  deploy-expo-preview:
+    name: deploy expo preview
+    if: github.event_name == 'pull_request'
+    needs: [build]
+    uses: nearform/optic-expo/.github/workflows/deploy-expo-preview.yml@chore/expo-preview
+    secrets:
+      expo-token: ${{ secrets.EXPO_TOKEN }}
```
simoneb, line 34: nit: add new line to avoid this red mark

simoneb (commented):
This works but I think we can do better. There is quite a bit of duplication between the existing publish workflow and this new reusable workflow.

They are doing exactly the same thing except for the handling of the release channel. I would suggest that:

1. we create a single reusable workflow which does the publishing
2. it takes as **input** the release channel, with the default value being default
3. it's fine to always comment on the PR even when the release channel is default
4. invoke this reusable workflow both from the publish workflow and from the ci workflow. In the latter case, it will provide a release-channel input with value `pr-${{ github.event.number }}`

Minor:

- I prefer removing all the names on the steps, they're usually redundant and unnecessarily clutter the workflow code
- point to specific versions of third party actions instead of master

allforabit:
I think it's right now and hope it's what you had in mind as far as sharing the expo publishing action. One thing is it's set to use this current branch ( `uses: nearform/optic-expo/.github/workflows/expo-publish.yml@chore/expo-preview`) but should be master. The build will fail if I do this though.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,15 @@
+name: publish
```
simoneb, line 1: let's keep the name consistent with the file name, hence: cd

.github/workflows/cd.yml
```diff
@@ -0,0 +1,15 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  publish:
+    name: publish to expo
+    uses: nearform/optic-expo/.github/workflows/expo-publish.yml@chore/expo-preview
+    with:
+      channel: default
```
simoneb, line 13: minor: shall we make this the default in the reusable workflow so if it's not specified it goes there?

.github/workflows/expo-publish.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  deploy-expo-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up repository
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
```
simoneb, line 23:
```suggestion
          node-version: 14
```

.github/workflows/expo-publish.yml
```diff
@@ -0,0 +1,42 @@
...
+        type: string
+
+jobs:
+  deploy-expo-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up repository
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v1
```
simoneb, line 21: please remove names. use latest versions of these actions

.github/workflows/expo-publish.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          node-version: 14.x
+      - name: Set up Expo
```
simoneb, line 24: remove name

.github/workflows/expo-publish.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: cp .env.sample .env
+      - name: Publish to Expo
+        run: expo publish --release-channel=${{ inputs.channel }}
+      - name: Pr Comment deployment link
+        if: github.event_name == 'pull_request'
+        uses: unsplash/comment-on-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          msg: App is ready for review, you can [see it here](https://expo.dev/@nearform/optic-expo?release-channel=pr-${{ github.event.number }}).
```
simoneb, line 42: remove all names, they're unnecessary clutter

.github/workflows/cd.yml
simoneb, line 1: since you're removing the publish workflow, change the readme to include this one please

simoneb (commented):
Some comments but this is the implementation I had in mind 👌

allforabit:
I've set this to run on the action from the master branch so it's failing. It'll work once it's merged in.

.github/workflows/expo-publish.yml
```diff
@@ -0,0 +1,36 @@
...
+          expo-token: ${{ secrets.expo-token }}
+      - run: yarn install
+      - run: cp .env.sample .env
+      - run: expo publish --release-channel=${{ inputs.channel }}
+      - uses: unsplash/comment-on-pr@master
+        if: github.event_name == 'pull_request'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          msg: App is ready for review, you can [see it here](https://expo.dev/@nearform/optic-expo?release-channel=pr-${{ github.event.number }}).
```
simoneb, line 36:
```suggestion
          msg: App is ready for review, you can [see it here](https://expo.dev/@nearform/optic-expo?release-channel=${{ inputs.channel }}).
```

README.md
```diff
@@ -1,7 +1,7 @@
 # Optic Expo
 
 [![ci](https://github.com/nearform/optic-expo/actions/workflows/ci.yml/badge.svg)](https://github.com/nearform/optic-expo/actions/workflows/ci.yml)
-[![publish](https://github.com/nearform/optic-expo/actions/workflows/publish.yml/badge.svg)](https://github.com/nearform/optic-expo/actions/workflows/publish.yml)
+[![publish](https://github.com/nearform/optic-expo/actions/workflows/cd.yml/badge.svg)](https://github.com/nearform/optic-expo/actions/workflows/publish.yml)
```
simoneb, line 4:
```suggestion
[![cd](https://github.com/nearform/optic-expo/actions/workflows/cd.yml/badge.svg)](https://github.com/nearform/optic-expo/actions/workflows/cd.yml)
```

simoneb (changes requested):

simoneb (approved):

PR #222, opened 2021-11-16
feat: enhance OTP request flow in the app

allforabit:
This PR implements the new flows described in https://www.figma.com/file/xsPf6IIM9AevLN5gZlXM4q/Optic-(Copy

The primary additions to the app are as follows:
- Users can now create multiple tokens per secret
- Each token can be given a description
- When an OTP request comes in it includes the token being used and the description added to it

The list of new screens are as follows:
- Create token page
- Edit token page
- List tokens for a secret
- OTP request screen

This PR also includes an update of the react native testing library as well as a fix to make tests start more quickly (by not including `useFakeTimers` in the jest setup file. This upgrade allows tests tests to be async. Previously tests would timeout if this was attempted.

To trigger the OTP flow call the following endpoint: 
`https://optic-zf3votdk5a-ew.a.run.app/api/generate/<token>`

Below are screenshots of the new flow taken on a real device:

Screenshot_20211124-085538
Screenshot_20211124-085628
Screenshot_20211124-085639
Screenshot_20211124-085847
Screenshot_20211124-085900
Screenshot_20211124-085952
Screenshot_20211124-090056
Screenshot_20211124-090111
)
Screenshot_20211124-095237

PR #265, opened 2021-12-01
feat: Allow user to filter by token

### LINKED ISSUE #262: Factor token in to token filter on token list page
See comment from PR https://github.com/nearform/optic-expo/pull/222:

> If the filter result is zero, could we filter by token?
> 
> For example: I see a token on my CI, and I want to check that the description match because I don't remember if I have copied&pasted it from another setup

On the tokens list page the filter should filter by the contents of the token if there is no match for the description.

ja-bravo:
See comment from PR #222:
 
>  If the filter result is zero, could we filter by token?
>  For example: I see a token on my CI, and I want to check that the description match because I don't remember if I have copied&pasted it from another setup

This PR also disables the auto correct on the search input

Closes #262

### DIFF
--- a/src/screens/TokensListScreen.tsx
+++ b/src/screens/TokensListScreen.tsx
@@ -80,8 +80,10 @@ export const TokensListScreen = ({ route, navigation }: Props) => {
 
   const filteredTokens = useMemo(() => {
     if (search.length > 1) {
-      return tokens.filter(({ description }) =>
-        description.toLowerCase().includes(search.toLowerCase())
+      return tokens.filter(
+        ({ token, description }) =>
+          description.toLowerCase().includes(search.toLowerCase()) ||
+          token.toLowerCase().includes(search.toLowerCase())
       )
     } else {
       return tokens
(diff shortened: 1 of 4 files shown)


src/screens/TokensListScreen.tsx
```diff
@@ -80,8 +80,10 @@ export const TokensListScreen = ({ route, navigation }: Props) => {
 
   const filteredTokens = useMemo(() => {
     if (search.length > 1) {
-      return tokens.filter(({ description }) =>
-        description.toLowerCase().includes(search.toLowerCase())
+      return tokens.filter(
```
Eomm, line 83: We don't have a test for this file, could we create it?

Eomm (approved):

src/screens/TokensListScreen.tsx
```diff
@@ -80,8 +80,10 @@ export const TokensListScreen = ({ route, navigation }: Props) => {
...
+        ({ token, description }) =>
```
simoneb, line 84: is there a chance the description does not have a value?

ja-bravo: I don't think so, this line on the create screen token blocks the creation if it's empty. https://github.com/nearform/optic-expo/blob/cdd26e79d0b59f6ea0f81e75f11f358159be24a0/src/screens/CreateTokenScreen.tsx#L46

It could be just whitespace but that wouldn't affect the functionality

simoneb (approved):

Issue #295, opened 2021-12-20
use biometric authentication to approve requests


simoneb:

salmanm:
The way I see this feature in action is as follows, let me know if it sounds ok.

- On OTP request page we have Approve & Reject buttons.
- When user clicks on Approve, we'd show a modal with fingerprint icon asking to authenticate using sensor
- The above will happen only when 
  - The device has fingerprint sensor
  - Perhaps a flag in the settings page that lets user enable/disable this feature.

simoneb:
I think I would do it slightly differently. 

1. there's one first step which I'm not sure is mandatory which is to "enroll" into biometric auth. I'm not sure if it's necessary or if we can interact with the biometric reader regardless. In any case I do agree that we should probably let the user decide whether to use it or not, when available
2. when bio auth is available/enabled, the page we show no longer has the approve/reject buttons but only shows that the user can approve by scanning the fingerprint (or whatever bio auth they have). this would basically simplify the approval process instead of requiring both bio and button

salmanm:
As discussed, we'd still need the Approve/Reject buttons as the local-auth (biometric/face/pattern/pin/pass) is handled by the OS which is platform dependent (modal dialog on iOS, modal popup from bottom on Android). So we can't directly bring that up before user gets a chance to read the description on the OTP request page. 

Implemented in https://github.com/nearform/optic-expo/pull/297

PR #297, opened 2021-12-22
feat: biometric auth

salmanm:
Closes https://github.com/nearform/optic-expo/issues/295


simoneb (approved):
Works nicely! There's a small UI glitch in the settings page, the setting name and the toggle are not vertically aligned. Once that's fixed this is good to go

PR #363, opened 2022-01-31
chores: bumping react native paper version

tiagoheliob:
Bumping react-native-paper version where on `<TextInput />` and `<Text />` have added some new required props, updating snapshots too.

closes #351


simoneb (approved):
Interesting that they introduced this breaking change in a minor version, and also strange that they made those props mandatory, because they feel like they have sensible defaults.

simoneb:
@tiagoheliob you can merge this, thanks

PR #436, opened 2022-03-28
feat: track pending notifications

### LINKED ISSUE #420: Show OTP requests on application open
Using the android app, if I miss the notification because I wipe out unfortunately or for whatever reason, I can not see the pending OTP requests by opening the application.

Is it possible to track all the OTP requests received and the status (waiting, approved, rejected, timeout)?

castarco:
## Changes
- Adds a new "notifications" section, where all pending notifications are listed to be handled (rejected, accepted, or dismissed in case they already expired).
- The pending notifications are persisted.
- The push notifications are automatically dismissed if the OTP requests are handled directly via the notifications section instead of doing that through the notifications provided by the smartphone's interface.

## About the scope
- For now, we don't keep track of all previous accepted & rejected OTP requests (once we handle them, we discard this information). I believe this should be discussed with more people, as it's a more advanced feature resembling a sort of audit log, and we might want to have support from the API side.

## Shortcomings
- Right now, the "notifications bell" does not have a nice numerical counter, it just turns to red instead when we have pending notifications. Before the reviews finish, I'll try to discover how to do that with our components. If don't arrive in time, we could defer it to another smaller PR.

Fixes #420 .

### DIFF
--- a/src/context/PendingNotificationsContext.tsx
+++ b/src/context/PendingNotificationsContext.tsx
@@ -1,0 +1,54 @@
+import React, {
+  createContext,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react'
+
+import * as storage from '../lib/secure-storage'
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface PendingNotification {}
+
+const defaultPendingNotifications: PendingNotification[] = []
+
+const initialContext = {
+  pendingNotifications: defaultPendingNotifications,
+}
+
+const PendingNotificationsContext = createContext(initialContext)
+
+export const PendingNotificationsProvider = ({ children }) => {
+  const [pendingNotifications, setPendingNotifications] = useState(
+    defaultPendingNotifications
+  )
+
+  useEffect(() => {
+    async function fn() {
+      const s = await storage.getObject<PendingNotification[]>(
+        'pendingNotifications'
+      )
+      if (s) {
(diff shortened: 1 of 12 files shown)


src/context/PendingNotificationsContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+  useEffect,
+  useMemo,
+  useState,
+} from 'react'
+import { Notification, dismissNotificationAsync } from 'expo-notifications'
+
+import * as storage from '../lib/secure-storage'
+import { NotificationData } from '../types'
+
+const defaultPendingNotifications: Notification[] = []
```
rp4rk, line 14: Could we extend this type to have `NotificationData` nested in order to avoid casting to it constantly?

rp4rk (approved):
LGTM but I'd wait for someone else to approve perhaps as well :)

src/context/PendingNotificationsContext.tsx
castarco: That sounds like a good idea, it's a pity that `Notification` is not generic itself.

rp4rk (approved):

castarco:
@simoneb In principle this is done, but I'm not sure if there's anything pending from my side to fix the `deploy-expo-preview` failed action. It seems related to missing credentials, and I'm not sure that I should touch anything on that side.

castarco:
"Cloning" this PR to a new one coming from a "local" branch (instead of my fork).

PR #446, opened 2022-03-31
feat: track pending notifications

### LINKED ISSUE #420: Show OTP requests on application open

castarco:
## Changes
- Adds a new "notifications" section, where all pending notifications are listed to be handled (rejected, accepted, or dismissed in case they already expired).
- The pending notifications are persisted.
- The push notifications are automatically dismissed if the OTP requests are handled directly via the notifications section instead of doing that through the notifications provided by the smartphone's interface.

## About the scope
- For now, we don't keep track of all previous accepted & rejected OTP requests (once we handle them, we discard this information). I believe this should be discussed with more people, as it's a more advanced feature resembling a sort of audit log, and we might want to have support from the API side.

## Shortcomings
- Right now, the "notifications bell" does not have a nice numerical counter, it just turns to red instead when we have pending notifications. Before the reviews finish, I'll try to discover how to do that with our components. If don't arrive in time, we could defer it to another smaller PR.

Fixes #420 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "react-native-screens": "~3.10.1",
     "react-native-svg": "12.3.0",
     "react-native-web": "0.17.1",
+    "react-timeago": "^6.2.1",
     "url-otpauth": "^2.0.0",
     "utility-types": "^3.10.0",
     "uuid": "8.3.2"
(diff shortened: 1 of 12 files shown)


rp4rk (approved):
LGTM 👍

simoneb:
@rp4rk once the preview is deployed can you test it out perhaps?

simoneb (approved):
Overall LGTM. For the badge containing the number of notifications, did you look into this approach? https://github.com/callstack/react-native-paper/issues/1466

castarco:
@simoneb I discovered it late, thank you for the reference, I already prepared a different branch for it.

PR #558, opened 2022-06-27
Linting fixes

### LINKED ISSUE #553: Fix React linting warnings.
See https://github.com/nearform/optic-expo/pull/552/files

aledgjones:
closes #553

Added `react/no-unstable-nested-components` exception for allowAsProps as the current usage seems valid based on the api provided by `react-native-paper`.

### DIFF
--- a/src/context/PendingNotificationsContext.tsx
+++ b/src/context/PendingNotificationsContext.tsx
@@ -48,8 +48,7 @@ export const PendingNotificationsProvider = ({ children }) => {
       const notificationId = notification.request.content.data.uniqueId
       const notAddedYet =
         pendingNotifications.findIndex(
-          notification =>
-            notification.request.content.data.uniqueId === notificationId
+          item => item.request.content.data.uniqueId === notificationId
         ) === -1
 
       if (notAddedYet) {
--- a/src/context/SecretsContext.tsx
+++ b/src/context/SecretsContext.tsx
@@ -48,10 +48,10 @@ export const SecretsProvider: React.FC<SecretsProviderProps> = ({
     let mounted = true
     async function initialize() {
       if (user) {
-        const secrets = await secretsManager.getAllByUser(user.uid)
+        const items = await secretsManager.getAllByUser(user.uid)
 
         if (mounted) {
-          setSecrets(secrets)
+          setSecrets(items)
         }
       }
     }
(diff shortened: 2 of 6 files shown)


simoneb:
what's the status of this PR?

src/context/PendingNotificationsContext.tsx
```diff
@@ -48,8 +48,7 @@ export const PendingNotificationsProvider = ({ children }) => {
       const notificationId = notification.request.content.data.uniqueId
       const notAddedYet =
         pendingNotifications.findIndex(
-          notification =>
-            notification.request.content.data.uniqueId === notificationId
+          item => item.request.content.data.uniqueId === notificationId
```
simoneb, line 51: why this rename?

src/context/SecretsContext.tsx
```diff
@@ -48,10 +48,10 @@ export const SecretsProvider: React.FC<SecretsProviderProps> = ({
     let mounted = true
     async function initialize() {
       if (user) {
-        const secrets = await secretsManager.getAllByUser(user.uid)
+        const items = await secretsManager.getAllByUser(user.uid)
```
simoneb, line 51: why this rename?

src/screens/TokenScreen.tsx
```diff
@@ -176,19 +175,29 @@ export const TokenScreen = ({ route, navigation }: Props) => {
   // Keep the description for the token up to date
   useEffect(() => {
     const updateDescription = async () => {
-      const tokens = secret.tokens ? [...secret.tokens] : []
-      const existingItemIndex = tokens.findIndex(item => item.token === token)
-      if (existingItemIndex === -1) {
+      const existingTokens = secret.tokens ? [...secret.tokens] : []
```
simoneb, line 178: I am not convinced that the new implementation is better than the original. In any case, let's please keep this PR to what the original intent is without introducing additional unrelated changes

simoneb (changes requested):

src/context/PendingNotificationsContext.tsx
aledgjones: Notification is already defined. I'm happy to update to a better name -- I just couldn't think of one!

simoneb: generally speaking shadowing variables is not a linting error, is it in this case? if so, fine with this change

aledgjones: Yes, it looks like these are showing as linting warnings in the link attached to the issue.

simoneb: okay fair enough

src/screens/TokenScreen.tsx
aledgjones: Understood, I'll revert these changes.

simoneb:
@aledgjones please take out of draft when ready

PR #613, opened 2022-07-27
fix: update unit tests to remove usage of mocked from ts-jest/utils

JamesLMilner:
Resolves the issues with upgrading to ts-jest@28.0.7  that you see in the action logs for https://github.com/nearform/optic-expo/pull/599

Main resolution is is removal of ts-jest/utils mocked as this is now built into jest (see https://github.com/kulshekhar/ts-jest/issues/3327)

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "jest-expo": "^44.0.0",
     "lint-staged": ">=13",
     "prettier": "^2.7.1",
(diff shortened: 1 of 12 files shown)


simoneb (approved):
LGTM, nice!

simoneb:
@JamesLMilner just some linter issues to fix

Issue #612, opened 2022-07-27
Numerous React tests log warnings about not using act with state updates


JamesLMilner:
Numerous tests log errors about not using act API around state updates with the tests. This should probably be remedied.

Here's an example in the actions logs: https://github.com/nearform/optic-expo/runs/7536227801?check_suite_focus=true#step:7:12

Here's 

```shell
console.error
      Warning: An update to TokenScreen inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at TokenScreen (/home/runner/work/optic-expo/optic-expo/src/screens/TokenScreen.tsx:77:31)
          at ThemeProvider (/home/runner/work/optic-expo/optic-expo/node_modules/@callstack/react-theme-provider/src/createThemeProvider.js:[14](https://github.com/nearform/optic-expo/runs/7536227801?check_suite_focus=true#step:7:15):3)
          at View
          at Component (/home/runner/work/optic-expo/optic-expo/node_modules/react-native/jest/mockComponent.js:28:18)
          at PortalHost (/home/runner/work/optic-expo/optic-expo/node_modules/react-native-paper/lib/commonjs/components/Portal/PortalHost.tsx:44:64)
          at Provider (/home/runner/work/optic-expo/optic-expo/node_modules/react-native-paper/lib/commonjs/core/Provider.tsx:27:11)

      [15](https://github.com/nearform/optic-expo/runs/7536227801?check_suite_focus=true#step:7:16)6 |     }
      157 |
    > 158 |     setIsRefreshing(false)
          |     ^
      159 |   }
      [16](https://github.com/nearform/optic-expo/runs/7536227801?check_suite_focus=true#step:7:17)0 |
      161 |   useEffect(() => {
```

PR #626, opened 2022-08-10
Fix: Warning texts in tests.

Thurfjell:
Updated tests for token screen components to not produce error logs. #612


simoneb (approved):
Warnings are gone, so this looks good 👌  Please link the issue though

PR #674, opened 2022-09-09
remove optional github token

marco-ippolito:
removing the optional dependabot github token

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,25 +1,23 @@
 name: ci
-
 on:
   push:
-    branches: [master]
+    branches:
+      - master
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -28,10 +26,6 @@ jobs:
       contents: write
     steps:
       - uses: fastify/github-action-merge-dependabot@v3
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          approve-only: true
```
simoneb, line 33: @marco-ippolito careful here, it removed too much! approve-only must stay

simoneb (approved):

Issue #591, opened 2022-07-12
Encrypt exported tokens file


davideroffo:
With this PR https://github.com/nearform/optic-expo/pull/590 the export and import feature has been added to the application.

Currently, we are storing a plain text file with the file content (information about secrets) easily readable and importable from any user.

We want to store the exported file with encryption, where the secret should be a concatenation of _[secret_string].[user_id]_
In this way, the file is no longer readable and can be imported only by the user (user_id) who created (exported) that file.

Things to do:
- [x] while exporting encrypt the file-content with _[secret_string].[user_id]_ secret
- [x] update the import method in order to decrypt first the file-content the user is trying to import
- [x] in order to be backward compatible, if a file is in plain text import it anyway for now (@simoneb is this ok?)

PR #688, opened 2022-09-14
Added the encryption to file exporting

SirCQQ:
Added the encryption to file exporting but kept plain text import still available for backward compatibility. 
Linked Issue https://github.com/nearform/optic-expo/issues/591

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,25 @@
 1. `yarn start`
 
 ## Development
+
 In order to successfully run the Optic-expo app locally you will need the following:
+
 1. Expo user account. You can sign up [here](https://expo.dev/signup).
 1. Once you have an Expo account, your account needs to be added to the NearForm organization (ask @simoneb to do that for you).
 1. Scan the QR code on your terminal or go to `exp://172.22.22.56:19000`
 1. Once the app loads, if you get the signin screen, in your terminal run `expo login -u <username> -p <password>` and reload the app
 
-* Note: The app doesn't run on the web
+## Notes
+
+- The app doesn't run on the web
+
+- `yarn start:native` is different from `yarn start` having the flag `--dev-client` you need to build a native version of the app as it generates a url like this:
+  `com.nearform.optic://expo-development-client/?url=http%3A%2F%2F192.168.5.92%3A8081` so you have to run `expo run:android` first.
+  exp://192.168.5.92:19000 </br> <br> As most things, it's a trade-off.Pp Without --dev-client it's faster to start working as you don't have to install native tools but it hides some problems that only occur in the native versions like the google login problem and the qr code scan.
 
 ## Figma Design
 
-The figma designs can be found [here](https://www.figma.com/file/xsPf6IIM9AevLN5gZlXM4q/Optic-(Copy))
+The figma designs can be found [here](<https://www.figma.com/file/xsPf6IIM9AevLN5gZlXM4q/Optic-(Copy)>)
 
 ## Sequence diagram
 
@@ -59,5 +67,6 @@ As you can see by the home page layout, you can generate many tokens to read the
 By doing so, you will be able to understand who is the caller that wants to read your OTP.
 
 Few notes on the NPM secret:
+
 - the NPM secret (QR or the textual code) is the one that you get, when you enable 2FA in your [npm](https://www.npmjs.com/) profile
 - if you already have 2FA activated (with a different authenticatior app e.g. Google Authenticator) and you need to get the secret, the easiest way is to get is to open the authenticator app you were using previously and get the QR code from that app and scan it with Optic Expo
--- a/app.config.js
+++ b/app.config.js
@@ -18,6 +18,7 @@ const config = {
     '230076165693-v8kk0ase89jp3s1qe0rtg70t3vb2h9i9.apps.googleusercontent.com',
   EAS_PROJECT_ID: 'e3c6e759-c8c5-41a8-861e-b3e4adab619a',
   APP_PACKAGE_ID: 'com.nearform.optic',
+  SECRET_STRING: 'kYfgT08ExAOKraIp7zNI',
 }
 
 export default {
@@ -43,6 +44,7 @@ export default {
     eas: {
       projectId: config.EAS_PROJECT_ID,
     },
+    secretString: config.SECRET_STRING,
   },
   splash: {
     image: './src/assets/splash.png',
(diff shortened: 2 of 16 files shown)


app.config.js
```diff
@@ -43,6 +44,7 @@ export default {
     eas: {
       projectId: config.EAS_PROJECT_ID,
     },
+    secretString: config.SECRET_STRING,
```
simoneb, line 47: this can't be hardcoded, the secret must be chosen by the user when exporting the secrets, so that only the user knows what is the password used to export them

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.4",
   "versionCode": 4,
   "scripts": {
-    "start": "expo start --dev-client",
+    "start": "expo start",
```
simoneb, line 7: @confraria does this change make sense to you? @SirCQQ said without this the app wouldn't start locally

src/lib/importExport.ts
```diff
@@ -7,11 +7,21 @@ import {
 } from 'expo-file-system'
 import * as Sharing from 'expo-sharing'
 import { Alert, Platform } from 'react-native'
+import CryptoJS from 'react-native-crypto-js'
 
 import { Secret } from '../types'
 
 const mimeType = 'application/octet-stream'
 
+function IsJsonString(str) {
```
simoneb, line 16: functions are camelCase in javascript

simoneb (changes requested):

package.json
confraria: The difference is that with `--dev-client` you need to build a native version of the app as it generates a url like this:
`com.nearform.optic://expo-development-client/?url=http%3A%2F%2F192.168.5.92%3A8081` so you have to run `expo run:android` first.

Without `--dev-client` you can just use expo go as the url it creates is `exp://192.168.5.92:19000`

As most things, it's a trade-off. Without `--dev-client` it's faster to start working as you don't have to install native tools but it hides some problems that only occur in the native versions like the google login problem and the qr code scan. 

We can remove it, add a `start:native` and update the README to reference these differences

package.json
```diff
@@ -57,6 +57,7 @@
     "react-dom": "18.0.0",
     "react-native": "0.69.5",
     "react-native-countdown-circle-timer": "^3.0.9",
+    "react-native-crypto-js": "^1.0.0",
```
confraria, line 60: As we are already using `aes-js` in the code (see `src/lib/secure-storage/aes.ts`) maybe we could avoid adding another dependency.

package.json
simoneb: thanks for the explanation. let's use as default what's most convenient for local development please, and change the readme accordingly

package.json
SirCQQ: After some investigation and testing, with the current implementation of the aes-js we would need to install another package anyway (scrypt or bcrypt) to convert a `string` to a valid `key` for aes-js encryption function, so we would end up exactly the same, the current implementation it's pretty straight forward and the package is made for  RN. Also, the package has the AES algorithm as well if we need it. (This is also based on a discussion with @simoneb)

confraria (approved):

src/lib/importExport.ts
```diff
@@ -7,11 +7,21 @@ import {
...
 
 const mimeType = 'application/octet-stream'
 
+function isJsonString(str) {
+  try {
+    JSON.parse(str)
+  } catch (e) {
+    return false
+  }
+  return true
```
simoneb, line 22: it would be more idiomatic to see this done as try { return true } catch { return false }

confraria:
tested export encrypted ✅ (confirmed the file is encrypted in adb cli)
tested import encrypted ✅ 
tested import plain text ✅
tested import plain text with a different uid ✅

src/screens/ImportFileSecret.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const handleDecryptImport = async () => {
+    setLoading(true)
+
+    try {
+      if (!secret.trim()) {
+        throw new Error('Secret cannot be empty')
+      }
+      const secrets = decryptDataToSecrets(
+        fileContent + ' ',
```
simoneb, line 55: why are you appending a string?

simoneb (changes requested):
See inline comments, plus, there are no tests for the feature. Please add tests

README.md
```diff
@@ -17,17 +17,26 @@
...
+
+- Note
+  `yarn start:narive` is different from `yarn start` having the flag `--dev-client` you need to build a native version of the app as it generates a url like this:
+  `com.nearform.optic://expo-development-client/?url=http%3A%2F%2F192.168.5.92%3A8081` so you have to run `expo run:android` first.
+  exp://192.168.5.92:19000 </br> </br> As most things, it's a trade-off. Without --dev-client it's faster to start working as you don't have to install native tools but it hides some problems that only occur in the native versions like the google login problem and the qr code scan.
 
 ## Figma Design
 
-The figma designs can be found [here](https://www.figma.com/file/xsPf6IIM9AevLN5gZlXM4q/Optic-(Copy))
+The figma designs can be found [here](<https://www.figma.com/file/xsPf6IIM9AevLN5gZlXM4q/Optic-(Copy)>)
```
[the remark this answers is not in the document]
SirCQQ: This was made by my editor on save ( I have formatOnSave on ) this is not breaking anything from what I see

src/lib/importExport.ts
```diff
@@ -67,14 +77,21 @@ export const doExport = (fileName, fileContent) => {
...
     return []
   }
-
   try {
     const result = await readAsStringAsync(uri)
-    const parsedSecrets = JSON.parse(result) as Secret[]
+    let confirmedString: string
+
+    if (isJsonString(result)) {
+      confirmedString = result
```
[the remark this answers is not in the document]
SirCQQ: This function is no longer used - deleted in the latest commit

simoneb:
@SirCQQ where do we stand with this PR?

SirCQQ:
@simoneb I forgot to push the changes for the comments and the tests on Friday, can you check if those are good?

app.json
```diff
@@ -0,0 +1,3 @@
+{
```
simoneb, line 1: why this file?

SirCQQ: Removed

simoneb (approved):

PR #722, opened 2022-09-22
fix: opening updates in expo go in android

### LINKED ISSUE #719: Authentication throws an error with expo go updates in android
error-login

https://github.com/expo/expo/blob/main/packages/expo-linking/src/Linking.ts#L158

expo-auth-session does not include the slash in the redirect URI and when we come back to the app it can't parse the update id from the url.

Problem can be solved here: https://github.com/expo/expo/blob/main/packages/expo-linking/src/Linking.ts#L158

There is already an issue related to this https://github.com/expo/expo/issues/16834

confraria:
fixes #719

### DIFF
--- a/patch/Linking.js
+++ b/patch/Linking.js
@@ -1,0 +1,342 @@
+import Constants from 'expo-constants';
+import { Platform, UnavailabilityError } from 'expo-modules-core';
+import invariant from 'invariant';
+import qs from 'qs';
+import { useEffect, useState } from 'react';
+import URL from 'url-parse';
+import NativeLinking from './ExpoLinking';
+import { hasCustomScheme, resolveScheme } from './Schemes';
+function validateURL(url) {
+    invariant(typeof url === 'string', 'Invalid URL: should be a string. Was: ' + url);
+    invariant(url, 'Invalid URL: cannot be empty');
+}
+function getHostUri() {
+    if (Constants.manifest?.hostUri) {
+        return Constants.manifest.hostUri;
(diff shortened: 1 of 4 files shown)


simoneb (approved):
Let's open an issue to undo this patching linking to the original issue, so we can monitor it

confraria:
solved the previous error, but we get a new one. 
Screenshot_20220922-170045
Screenshot_20220922-170057

patch/Linking.js
```diff
@@ -0,0 +1,341 @@
...
+        return removeScheme(Constants.linkingUri).replace(/\/--($|\/.*$)/, '');
+    }
+    else {
+        return null;
+    }
+}
+function isExpoHosted() {
+    const hostUri = getHostUri();
+    return !!(hostUri &&
+        (/^(.*\.)?(expo\.io|exp\.host|exp\.direct|expo\.test|expo\.dev)(:.*)?(\/.*)?$/.test(hostUri) ||
```
confraria, line 32: `|expo\.dev` was missing from this regex 
🙃 Keith from expo's discord channel gave me the tip!

simoneb:
does expo have an open issue about this? can we contribute this back to them perhaps? in any case, let's track an issue in this repo to remind ourselves to undo this workaround

confraria:
- tested this with both native apps running on ios and android ✅ and no regresssion occured
- Draft PR opened in expo repo and issue added here to remove the patch ✅

PR #724, opened 2022-09-23
chore: change ci to only build on release branch automation and submit to app store on release completed

### LINKED ISSUE #683: Reconsider if building the apps on each post to master is needed
At the moment the CI is building the apps with preview profile on each push to master.
As the production build and subsequent submission to stores never goes directly to users (there is a test phase in both app store and play store) this might not be needed.

Reconsider the utility of it after a few releases.

confraria:
closes #683

### DIFF
--- a/.github/workflows/expo-eas-build.yml
+++ b/.github/workflows/expo-eas-build.yml
@@ -4,7 +4,7 @@ on:
   workflow_call:
     secrets:
       expo-token:
-        description: 'expo token to use to publish preview'
+        description: 'expo token'
         required: true
     inputs:
       platform:
@@ -28,7 +28,8 @@ jobs:
       - uses: expo/expo-github-action@v7
         with:
           expo-cache: true
-          expo-version: 5.x
+          expo-version: latest
+          eas-version: latest
           token: ${{ secrets.expo-token }}
       - run: yarn
-      - run: npx eas build --platform ${{ inputs.platform }} --non-interactive --profile=${{ inputs.profile }} --no-wait ${{ inputs.submit && '--auto-submit' ||  ''}}
+      - run: npx eas build --platform ${{ inputs.platform }} --non-interactive --profile=${{ inputs.profile }} ${{ inputs.submit && '--auto-submit' ||  ''}}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -12,20 +12,16 @@ jobs:
     secrets:
       expo-token: ${{ secrets.EXPO_TOKEN }}
 
-  build-ios:
-    uses: ./.github/workflows/expo-eas-build.yml
+  submit-ios:
+    uses: ./.github/workflows/expo-eas-submit.yml
     with:
-      profile: production
-      platform: ios
-      submit: true
+      platform: 'ios'
     secrets:
       expo-token: ${{ secrets.EXPO_TOKEN }}
 
-  build-android:
-    uses: ./.github/workflows/expo-eas-build.yml
+  submit-android:
+    uses: ./.github/workflows/expo-eas-submit.yml
     with:
-      profile: production
-      platform: android
-      submit: true
+      platform: 'android'
     secrets:
       expo-token: ${{ secrets.EXPO_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,20 +42,20 @@ jobs:
 
   build-ios:
     needs: build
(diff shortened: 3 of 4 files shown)


confraria:
⚠️ the title of the PR is just to test the action

confraria:
Builds are triggered when the PR has the name of the optic release branch ✅
Screenshot 2022-09-23 at 12 22 05

confraria:
eas-cli was removed from the package.json because of this recommendation:
https://github.com/expo/eas-cli#enforcing-eas-cli-version-for-your-project

> Installing eas-cli to your project dependencies is strongly discouraged because it may cause dependency conflicts that are difficult to debug.

.github/workflows/expo-eas-clean.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: expo/expo-github-action@v7
+        with:
+          expo-cache: true
+          expo-version: latest
+          eas-version: latest
+          token: ${{ secrets.expo-token }}
+      - run: eas channel:delete ${{ format('pr-{0}', github.event.number) }} --non-interactive
+      - run: eas branch:delete ${{ format('pr-{0}', github.event.number) }} --json
```
confraria, line 18: in this command --non-interactive doesn't work. but --json triggers the non-interactive mode

package.json
```diff
@@ -32,7 +32,6 @@
     "aes-js": "^3.1.2",
     "buffer": "^6.0.3",
     "dotenv": "^16.0.2",
-    "eas-cli": "^2.1.0",
```
confraria, line 35: due to this recommendation:
https://github.com/expo/eas-cli#enforcing-eas-cli-version-for-your-project
>Installing eas-cli to your project dependencies is strongly discouraged because it may cause dependency conflicts that are difficult to debug.

.github/workflows/expo-eas-clean.yml
```diff
@@ -0,0 +1,18 @@
+name: expo-eas-clean
```
confraria, line 1: to clean the eas update list after the PR is closed.

simoneb: I thought this was done automatically!

.github/workflows/expo-eas-build.yml
```diff
@@ -28,7 +28,8 @@ jobs:
       - uses: expo/expo-github-action@v7
         with:
           expo-cache: true
-          expo-version: 5.x
+          expo-version: latest
+          eas-version: latest
           token: ${{ secrets.expo-token }}
       - run: yarn
-      - run: npx eas build --platform ${{ inputs.platform }} --non-interactive --profile=${{ inputs.profile }} --no-wait ${{ inputs.submit && '--auto-submit' ||  ''}}
+      - run: eas build --platform ${{ inputs.platform }} --non-interactive --profile=${{ inputs.profile }} ${{ inputs.submit && '--auto-submit' ||  ''}}
```
simoneb, line 35: do we still need the extra input(s) such as `submit`?

.github/workflows/expo-eas-clean.yml
simoneb, line 18: dependabot PRs will not have channel/branches as they're explicitly excluded. are we sure these commands won't fail in such cases?

simoneb (commented):
See inline comments plus a general consideration.

I see we're now building the binaries based on the PR which the release automation opens automatically. I'm not entirely sure that's a good enough trigger. I mean that PR is meant to be something fairly transient, which in fact users could also close if they change their mind.

I wonder if:

- is there a real need to build and publish at different stages, rather than simply build AND publish together, but simply later than we were doing before?
- if the answer to the previous point is NO, can we identify a better trigger for build?

.github/workflows/expo-eas-clean.yml
confraria: you're right! added an if to exclude those

confraria:
> 

My line of thought was: avoid completing the release without being 100% sure the native builds are successful.
We could simply put the build and submission on the release trigger, but if the build failed then we would have to do another release.

.github/workflows/expo-eas-build.yml
confraria: they are not used now. I left them just in case we want to change the flow and use them in a different way.

simoneb (approved):

.github/workflows/expo-eas-clean.yml
simoneb: I'm fairly confident that before eas they were being deleted automatically, I mean by expo, not us

PR #771, opened 2022-10-07
fix: store cleaned up secret in secure storage

ifero:


src/screens/TypeScreen.tsx
```diff
@@ -43,7 +43,8 @@ export const TypeScreen: React.FC<TypeScreenProps> = ({ navigation }) => {
   const [account, setAccount] = useState(user.name || '')
 
   const handleAddSecretButtonPress = async () => {
-    await add({ uid: user.uid, secret, account, issuer })
+    const cleanedSecret = secret.trim().replace(/ /g, '').toUpperCase()
```
ilteoood, line 46: what if you remove the trim function and use `/\s/g` as replace regex?

mahenrique94: Yeah, you can remove `trim` in favor to use only `replace` with regex suggested by @ilteoood .

ifero: Will do. I've blindly copy-pasted 😅

src/screens/TypeScreen.test.tsx
```diff
@@ -55,8 +53,68 @@ describe('TypeScreen', () => {
...
+
+        fireEvent.changeText(getByLabelText('Issuer'), 'Issuer A')
+        fireEvent.changeText(getByLabelText('Secret'), 'mysecret!1')
+        fireEvent.changeText(getByLabelText('Account'), 'My Account')
+      })
+
+      it('should show an error message', () => {
+        const { getByText } = screen
+
+        expect(getByText('Invalid secret')).toBeDefined()
```
ilteoood, line 74: if the getBy fails, it will throw an exception so here the text is always defined.
I suggest to change the `getBy` with a `queryBy` or the check the `toBeInTheDocument` instead of `toBeDefined`

ilteoood (commented):
LGTM

src/screens/TypeScreen.test.tsx
ifero: Well, the meaning of this test is exactly that: if the text is not there, it should throw an exception and fail.
`queryBy` usually will return a false positive test in case the text is not visible at all, because it will not throw an exception but will return `null` which is interpreted as `defined` for some reason.

Regarding the `toBeInTheDocument` we can't use it here because is part of `@testing-library/jest-dom` which (imho) shouldn't be used in native apps.

If is ok I would stick to this implementation, but open to change my mind :)

ilteoood: For the queryBy, the check is just `receivedValue !== undefined`, so null is accepted. That's why you should change the check to `.not.toBeNull()`
BTW, this implementation is completely fine. My comment was just to give a better output if the test fails, instead of an exception 😄

ilteoood (approved):
LGTM

src/screens/TypeScreen.tsx
```diff
@@ -43,12 +43,13 @@ export const TypeScreen: React.FC<TypeScreenProps> = ({ navigation }) => {
...
   const handleAddSecretButtonPress = async () => {
-    await add({ uid: user.uid, secret, account, issuer })
+    const cleanedSecret = secret.replace(/\s/g, '').toUpperCase()
+    await add({ uid: user.uid, secret: cleanedSecret, account, issuer })
     navigation.navigate('Home')
   }
 
   const invalidSecret = !RFC4648_REGEX.test(
-    secret.trim().replace(/ /g, '').toUpperCase()
+    secret.replace(/\s/g, '').toUpperCase()
```
salmanm, line 52: why not use `cleanedSecret` here?

ilteoood: Because it is out of scope, as it is defined in the `handleAddSecretButtonPress`
You could always extract it at the same level of the `useState` and reuse it everywhere

src/screens/TypeScreen.test.tsx
ifero: Make sense, to be fair 😄 
Will take it under consideration next time

src/screens/TypeScreen.tsx
ifero: just because `cleanedSecred` is not in the same scope. 
I think we can achieve something more efficient here, let me try something 😉

src/screens/TypeScreen.test.tsx
```diff
@@ -55,8 +53,68 @@ describe('TypeScreen', () => {
...
+      beforeEach(() => {
+        const { getByLabelText } = setup()
+
+        fireEvent.changeText(getByLabelText('Issuer'), 'Issuer A')
+        fireEvent.changeText(getByLabelText('Secret'), 'mysecret!1')
+        fireEvent.changeText(getByLabelText('Account'), 'My Account')
+      })
+
+      it('should show an error message', () => {
+        const { getByText } = screen
```
simoneb, line 72: folks let's try to be consistent with existing code too. either use screen or use the return value of setup() but avoid using both as it gets quite confusing

PR #555, opened 2022-06-23
upgrade expo to 45, remove deprecated expo-app-loading

aledgjones:
Closes #521

expo-app-loading is deprecated and as suggested in the deprecation notice this PR includes a migration to use expo-splash-screen directly

PR #777, opened 2022-10-07
775 use expo splash screen instead of expo app loading

### LINKED ISSUE #775: Use [expo-splash-screen] instead of [expo-app-loading]
`expo-app-loading` has been deprecated in favor of `expo-splash-screen`

image

ifero:
Closes #775

### DIFF
--- a/src/Main.tsx
+++ b/src/Main.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { NavigationContainer } from '@react-navigation/native'
 import { createStackNavigator } from '@react-navigation/stack'
 import {
@@ -13,7 +13,7 @@ import {
   useFonts as useFiraCode,
   FiraCode_400Regular,
 } from '@expo-google-fonts/fira-code'
-import AppLoading from 'expo-app-loading'
+import * as SplashScren from 'expo-splash-screen'
 import { NativeStackScreenProps } from 'react-native-screens/native-stack'
 
 import theme from './lib/theme'
@@ -63,6 +63,8 @@ export type MainStackParamList = {
   }
 }
 
+SplashScren.preventAutoHideAsync()
+
 export default function Main() {
   const [hasPoppinsLoaded] = usePoppins({
     Poppins_700Bold,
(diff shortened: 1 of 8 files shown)


src/Main.tsx
```diff
@@ -76,10 +78,16 @@ export default function Main() {
...
   })
 
-  const { user } = useAuth()
+  const { user, loading } = useAuth()
+
+  useEffect(() => {
+    if (hasPoppinsLoaded && hasDidactLoaded && hasFiraCodeLoaded && !loading) {
+      SplashScren.hideAsync()
+    }
+  }, [hasDidactLoaded, hasFiraCodeLoaded, hasPoppinsLoaded, loading])
```
ilteoood, line 87: as it depends on so many fields, I think that the effect can be removed

simoneb:
I think this was attempted before, can you check if there's anything valuable in the previous attempt? https://github.com/nearform/optic-expo/pull/555/files#diff-1d505bc481460af8fd3c004784c3320e754ac12542df57c4309dbe17ac781f42

ifero:
> I think this was attempted before, can you check if there's anything valuable in the previous attempt? https://github.com/nearform/optic-expo/pull/555/files#diff-1d505bc481460af8fd3c004784c3320e754ac12542df57c4309dbe17ac781f42

@simoneb Conceptually speaking, is a good solution. But, I guess, it may end up in a situation in which the `SplashScreen.hideAsync()` will not be fired.

My assumptions are based on:
```
onLayout ​
Invoked on mount and on layout changes. This event is fired immediately once the layout has been calculated, but the new layout may not yet be reflected on the screen at the time the event is received, especially if a layout animation is in progress.
```

we aim to call
```
useCallback(async () => {
    if (isReady) {
      await SplashScreen.hideAsync()
    }
  }, [isReady])
```
through `onLayout` which may happen before `isReady` is true. In this case (I SUPPOSE, can't be 100% sure) we may never remove the SplashScreen even tho the app is ready to be consumed.

Happy to hear @ilteoood and @confraria voice here too

src/Main.tsx
ifero: make sense :)

simoneb:
@ifero ok sounds good. I tried the preview build but I can never get the splashscreen to show. What do you think about doing a further improvement here?

1. let's introduce a fake delay so that the splashscreen always appears when you open the app. I think that having a consistent flow of screens would help user experience
2. there's already a race condition of some sort because when you open the app, sometimes you're shown the login screen for a short time then the app figures that you're already logged in and takes you to the main screen. this should not happen. how about we show the splash screen until we know for sure that you're either logged in or logged out, and only then proceed to either the main screen or the login screen, respectively?

ifero:
@simoneb I've tried previews build too (both my branch and main branch) and on device it never appears while on simulators it appear.

It may be an issue with expo go, but not sure.

regarding (2.) I've added a check to wait until the "loading" of the Auth check is done and should avoid showing the authentication screen

simoneb (approved):

PR #839, opened 2022-11-15
feat: display otp notifications with sound when app is already open

### LINKED ISSUE #828: When a request arrives when the app is open, no notification happens
When you receive a request for generating a OTP when the app is open, the bell icon gets a badge but no notification happens, meaning that unless you're looking at the phone's screen you don't know it happened.

Even when the app is open, the notification should be displayed in the same way as when the app is closed.

guilhermelimak:
Closes #828.


simoneb (approved):
Not entirely sure this will work fully because I would have expected to find a different configuration for the app when it's in the foreground and when it's in the background, or iow, even though they were false, both the sound and the badge (whatever it is) I believe were played with the app in the background, but not in the foreground. Anyway, this doesn't look like it'll do any damage so let's merge it and see what happens I guess

guilhermelimak:
> Not entirely sure this will work fully because I would have expected to find a different configuration for the app when it's in the foreground and when it's in the background, or iow, even though they were false, both the sound and the badge (whatever it is) I believe were played with the app in the background, but not in the foreground. Anyway, this doesn't look like it'll do any damage so let's merge it and see what happens I guess

From the [docs](https://docs.expo.dev/versions/latest/sdk/notifications/#setnotificationhandlerhandler-notificationhandler--null-void) it seems like the notification handler is only used when the app is running. I also tested a couple of times to make sure it worked so I'm pretty confident.

PR #840, opened 2022-11-15
feat: add app version to settings page

### LINKED ISSUE #832: Display application version in settings page

guilhermelimak:
This will not work properly when testing on Expo go since it returns the native app version and it will use Expo's version (2.26.4 on my phone) instead of Optic's which should be something like 1.0.7. 

#### Image preview

Closes #832.


simoneb (approved):
Ok, it would be great to have something that works everywhere but I guess it's not easy to do

PR #841, opened 2022-11-15
fix: show `add secret` button only on home screen

### LINKED ISSUE #831: The "+" (plus) button in the settings page should be removed
It doesn't make any sense for it to be there

guilhermelimak:
I removed the `portal` wrapping the `FAB.Group` component so it's rendered at the page where it's being used but according to the react-native-paper docs it could cause z-index issues. I did try a couple of different scenarios (long list of secrets scrolling under the button, trying to use the buttons and dropdowns while close to the FAB) and it all seemed to work fine.

If that's not an acceptable solution I can try to investigate deeper since the logic to show/hide it seems correct (it uses react-navigation's isFocused hook to check whether a page is focused and shows the button only if the homepage is focused).

Closes #831.

### DIFF
--- a/src/components/Actions.tsx
+++ b/src/components/Actions.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react'
 import { StyleSheet } from 'react-native'
-import { FAB, Portal } from 'react-native-paper'
+import { FAB } from 'react-native-paper'
 
 import theme from '../lib/theme'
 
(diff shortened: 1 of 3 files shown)


simoneb:
ok so the [build error](https://github.com/nearform/optic-expo/actions/runs/3481164029/jobs/5821944256#step:6:7) is caused by https://github.com/nearform/optic-expo/pull/844. we probably need to do some wrapping in quotes or something

simoneb (approved):
meanwhile approving to avoid blocking, but we should find a solution for the eas error

PR #844, opened 2022-11-16
fix: show version on eas-update for production release

### LINKED ISSUE #830: Fix no-message when calling eas-update for a production release
It would make sense to provide the release version instead

brunopascottini:
Closes #830


simoneb (commented):
shall we consider removing the default value of this input in the called workflow?

brunopascottini:
@simoneb Sounds good, I'll also remove from the other input and from the CI workflow.

.github/workflows/ci.yml
```diff
@@ -35,8 +35,8 @@ jobs:
     if: github.actor != 'dependabot[bot]'
     uses: ./.github/workflows/expo-eas-update.yml
     with:
-      channel: ${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.number) || 'preview'}}
```
simoneb, line 38: why did you remove this?

package.json
```diff
@@ -51,9 +51,9 @@
     "expo-random": "~13.0.0",
     "expo-secure-store": "~12.0.0",
     "expo-sharing": "~11.0.1",
-    "expo-splash-screen": "~0.17.4",
+    "expo-splash-screen": "~0.17.5",
     "expo-status-bar": "~1.4.2",
-    "expo-updates": "~0.15.4",
+    "expo-updates": "~0.15.5",
```
simoneb, line 56: why did  you update package versions?

brunopascottini: Build was failing on CI and I thought it would be better to fix now than leave it to dependabot.

.github/workflows/ci.yml
brunopascottini: I checked the previous commits that triggered this workflow and this default message was not being used.

package.json
simoneb: oh interesting. it should have been caught before being merged to master, that's weird!

simoneb (approved):

Issue #869, opened 2022-11-29
Fix expo doctor warnings


lasmil:
While running npm run doctor (or with yarn) we have seen these:
```
Some dependencies are incompatible with the installed expo package version:
 - expo-barcode-scanner - expected version: ~12.1.0 - actual version installed: 12.0.0
 - expo-camera - expected version: ~13.1.0 - actual version installed: 13.0.0
```
 
 Because of this, there is a build issue in the open PRs.

PR #870, opened 2022-11-29
fix: 869: fix expo doctor package version issues

lasmil:
- closes https://github.com/nearform/optic-expo/issues/869
- to be able to run the ci build, we need to able to run the  'yarn doctor' command
- this PR updates the problematic packages
- this should also fix the dependabot created PR build issues


simoneb:
@lasmil thanks for this. as part of this work I would like us to understand why this happens, as I've seen this happen in the past but I don't understand why.

- we run doctor in CI, meaning that no incompatible dependencies should be merged
- yet it looks like that for some reason incompatible dependencies still make their way into the master branch

Can we figure out why?

lasmil:
> @lasmil thanks for this. as part of this work I would like us to understand why this happens, as I've seen this happen in the past but I don't understand why.
> 
> * we run doctor in CI, meaning that no incompatible dependencies should be merged
> * yet it looks like that for some reason incompatible dependencies still make their way into the master branch
> 
> Can we figure out why?

I looked into what are the possible explanations for this. Other than manually adding specific versions into package.json, I can't seem to find out why this could have happened.
The last successful PR was the dependabot triggered [expo version update](https://github.com/nearform/optic-expo/pull/862). I checked out these changes and yarn doctor fails for me as well.

Maybe @SociableSteve can give an opinion (or suggest somebody who can help) on how can yarn doctor miss these dependency related problems.

Other than this, I would try to see if these changes (in this PR) can actually fix the other open PRs build problems and then keep an eye open to certain package.json changes (dependabot and not dependabot triggered). Maybe even manually running a yarn doctor command on these to try to catch the issue.

simoneb:
dependabot should be the only way that changes to package.json make their way to the main branch, so I guess we still don't know why this is happening.

simoneb (approved):

lasmil:
> dependabot should be the only way that changes to package.json make their way to the main branch, so I guess we still don't know why this is happening.

I will keep an eye open to see if I can spot this happening in future PRs. Maybe we can find something.

PR #1012, opened 2023-02-21
feat: switch preview deploy action

### LINKED ISSUE #1006: Use official expo action for deploy previews when published
Currently we're using a custom approach built in a workflow. When the official support is released, we should switch to that one

https://github.com/expo/expo-github-action/pull/206

davideroffo:
Closes #1006 

WIP


simoneb:
Closing as this action is now deprecated and we're going to be switching to the new action (which we're already using for building the prod version of the app)

PR #1095, opened 2023-04-05
chore: update dependabot ignore list

### LINKED ISSUE #1090: Update dependabot ignore list
We currently ignore all `expo-*` packages: https://github.com/nearform/optic-expo/blob/master/.github/dependabot.yml#LL8C26-L8C32

`react` and `react-native` updates should be managed when updating expo with its cli and should also be ignored

`jest-expo` is similar to the `expo-*` packages and should be manually updated when upgrading the expo version (via expo doctor)

anapaulalemos:
Closes #1090

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
       interval: 'daily'
     ignore:
       - dependency-name: 'expo-*'
+      - dependency-name: 'react*'
+      - dependency-name: 'jest-expo'
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:


.github/dependabot.yml
```diff
@@ -6,6 +6,8 @@ updates:
       interval: 'daily'
     ignore:
       - dependency-name: 'expo-*'
+      - dependency-name: 'react*'
```
simoneb, line 9: this will catch several packages that won't be updated by expo https://github.com/nearform/optic-expo/blob/master/package.json#L63-L71. it's not a big deal in itself, but nobody will ever update those packages then, which is potentially an issue. this is the compromise that's difficult to accept, and one reason why we haven't done this yet.

as an alternative, and smarter approach, if we could figure out (without hardcoding) what packages expo wants to manage itself, we could build an action which automatically closes PR opened by dependabot on those packages.

anapaulalemos: Yeah, I added this general `react*` ignore inspired by the way it was done for the `expo-*` packages. But I understand and agree that it is not the best approach for the case. I'll close this PR and link your comment to the created issue, in order to give more context to it.

PR #1118, opened 2023-04-14
chore: close PR if dependabot is trying a bump that should be managed by Expo

### LINKED ISSUE #1090: Update dependabot ignore list

anapaulalemos:
Closes  #1090 

* Checks if the PR is a dependabot PR
* Checks which expo version we're running on
* Grabs the packages which are related to that version
* Check if it's bumping one of those packages
* Closes the PR as not planned in that case

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,68 @@ name: ci
 on:
   push:
     branches:
-      - master
+      - feat/update-ci-workflow
   pull_request:
     branches:
-      - master
+      - feat/update-ci-workflow
 jobs:
+  pre-build:
+  # if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+      - run: yarn
+
+      - name: Check Expo version
+        id: version
+        run: node -p -e '`expo-package-version=${require("./package.json").dependencies.expo}`' >> $GITHUB_OUTPUT
+      - uses: actions/github-script@v6
+        id: sanitise-package-version
+        with:
+          script: return `${{ steps.version.outputs.expo-package-version }}`.replaceAll('~','')
+          result-encoding: string
+
+      - name: Grab Expo related packages
+        uses: actions/github-script@v6
+        id: related-packages
+        with:
+          result-encoding: string
+          script: |
+            const expoVersion = `${{ steps.sanitise-package-version.outputs.result }}`
+            const majorExpoVersion = `${expoVersion.substring(0, expoVersion.indexOf('.'))}.0.0`
+
+            const expoData = await fetch('https://api.expo.dev/v2/versions/latest')
+            const expoDataOutput = await expoData.json()
+
+            const { relatedPackages } = expoDataOutput.data.sdkVersions[majorExpoVersion]
+            return relatedPackages
+
+      - name: Check if it is a bump that should be managed by Expo
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
+          script: |
+            const shouldClosePr = [
+              ...Object.keys(${{ steps.related-packages.outputs.result }}),
+              'react',
+              'react-native',
+              'expo'
+            ].some(dependency => ${{ github.event.pull_request.title }}.includes(dependency))
+
+            if(shouldClosePr) {
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: '🚫 This PR was closed because this dependency must be managed by the Expo itself'
+              })
+            }
+
   build:
+    needs: pre-build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


simoneb (changes requested):
nice job here, now let's extract this into a single js script that we invoke using the github-script function instead of orchestrating all this using the workflow, which makes it more complicated. 

as an alternative, we could keep this and extract it into a reusable workflow.

one thing that's missing is that if we close the PR, we need to find a way to stop the execution of the rest of the worfklow

.github/workflows/check-expo-dependencies.yml
```diff
@@ -0,0 +1,74 @@
...
+              await github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: '@dependabot close'
+              })
+
+              return 'true'
+            }
+            return 'false'
```
simoneb, line 72: this works, but can you see this is all JS code? why don't you extract it in a js file and then run it with the github-script action, instead of having to write js code in yaml file?

.github/workflows/ci.yml
```diff
@@ -7,9 +7,29 @@ on:
...
+
+  pre-build:
+    if: github.actor == 'dependabot[bot]'
+    name: Check expo dependencies
+    uses: ./.github/workflows/check-expo-dependencies.yml
+
   build:
     runs-on: ubuntu-latest
+    needs: pre-build
+    if: always()
```
simoneb, line 19: I don't think this if is needed

.github/workflows/ci.yml
```diff
@@ -7,9 +7,29 @@ on:
...
     steps:
+      - name: Cancel workflow
+        if: github.actor == 'dependabot[bot]' && needs.pre-build.outputs.cancel-workflow == 'true'
```
simoneb, line 22: do we need to check if the actor is dependabot?

.github/workflows/check-expo-dependencies.yml
```diff
@@ -0,0 +1,74 @@
...
+            const shouldClosePR = [
+              ...Object.keys(relatedPackages),
+              'react',
+              'react-native',
+              'expo' 
+            ].some(dependency => `${{ github.event.pull_request.title }}`.includes(dependency))
+
+            if (shouldClosePR) {
+              // ask dependabot to close the PR
+              await github.rest.issues.createComment({
```
simoneb, line 63: let's not ask dependabot to close the PR, let's just close it

.github/workflows/ci.yml
anapaulalemos: As the `build` job depends on the `pre-build` if the last one is not executed (if it is not a PR from dependabot, for example) it was not being executed either. So the `if: always()` ensures that the `build` job runs regardless of whether the `pre-build` happens or not.

.github/workflows/check-expo-dependencies.yml
```diff
@@ -0,0 +1,45 @@
...
+  check:
+    runs-on: ubuntu-latest
+    outputs:
+      cancel-workflow: ${{ steps.should-cancel-workflow.outputs.result }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+      - run: yarn
```
simoneb, line 19: why do we need to check out the source code and install packages again?

.github/workflows/check-expo-dependencies.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Check Expo version
+        id: version
+        run: node -p -e '`expo-package-version=${require("./package.json").dependencies.expo}`' >> $GITHUB_OUTPUT
```
simoneb, line 23: can't we move this inside the node script?

.github/workflows/check-expo-dependencies.yml
simoneb: try moving the usage of this workflow **after** the checkout action in the calling workflow :)

simoneb (changes requested):
Considering the limitations we have at the moment I would suggest that we get rid of the reusable workflow and include the execution of the script inside the main workflow, clearly after the checkout and setup-node actions. Let's also include in the script the WHOLE logic, including the cancelation of the workflow execution and closing of the PR, it can all be done within the script

.github/workflows/ci.yml
```diff
@@ -11,6 +11,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - name: Check if it is a dependency that should be managed by Expo
+        if: github.actor == 'dependabot[bot]'
```
simoneb, line 16: this shouldn't run if it's not a pull request, we need to put a condition in here

.github/workflows/ci.yml
```diff
@@ -11,6 +11,24 @@ jobs:
...
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
```
simoneb, line 19: we don't care about the result encoding any longer now

.github/workflows/ci.yml
```diff
@@ -11,6 +11,24 @@ jobs:
...
+          script: |
+            const script = require(
+              `${process.env.GITHUB_WORKSPACE}/scripts/check-expo-dependencies.js`
```
simoneb, line 22: I don't think prefixing this with the workspace env var is necessary. the working folder is already that where the code is checked out

scripts/check-expo-dependencies.js
```diff
@@ -0,0 +1,39 @@
+module.exports = async ({ context, github, fetch, prTitle }) => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const expoVersion = require('../package.json').dependencies.expo
+  const expoVersionSanitized = expoVersion.replaceAll('~', '')
```
simoneb, line 4: are you sure that tilde is the only possible symbol? maybe it is now, but maybe expo at some point may decide to use other version specifiers too, e.g. ^

scripts/check-expo-dependencies.js
```diff
@@ -0,0 +1,39 @@
...
+  )}.0.0`
+
+  const expoData = await fetch('https://api.expo.dev/v2/versions/latest')
+  const expoDataOutput = await expoData.json()
+
+  const { relatedPackages } = expoDataOutput.data.sdkVersions[majorExpoVersion]
+
+  const shouldClosePR = [
+    ...Object.keys(relatedPackages),
+    'react',
```
simoneb, line 18: this doesn't look very safe. I mean surely there are dependencies with the "react" string occurring in their name that we don't want to block. I'd try to be more specific about the dependency names

simoneb (changes requested):

scripts/check-expo-dependencies.js
anapaulalemos: Now I'm using `dependabot-metadata` which returns the list of dependency names being updated to match the related packages to avoid the problem you noticed above.

.github/workflows/ci.yml
```diff
@@ -11,6 +11,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - name: Dependabot metadata
+        if: github.actor == 'dependabot[bot]' && github.event_name == 'pull_request'
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.3.1
```
simoneb, line 18: use the major version specifier only please, as we do with all actions, to avoid too much noise from dependabot when trying to bump this version

scripts/check-expo-dependencies.js
```diff
@@ -0,0 +1,45 @@
+module.exports = async ({ context, github, fetch, dependencyNames }) => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const expoVersion = require('../package.json').dependencies.expo
+  const expoVersionSanitized = expoVersion
+    .replaceAll('~', '')
+    .replaceAll('^', '')
```
simoneb, line 6: minor. let's use a nice regexp to do this in one go

scripts/check-expo-dependencies.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const expoData = await fetch('https://api.expo.dev/v2/versions/latest')
+  const expoDataOutput = await expoData.json()
+
+  const { relatedPackages } = expoDataOutput.data.sdkVersions[majorExpoVersion]
+  const dependencyNamesList = dependencyNames.split(',')
+
+  const shouldClosePR = [
+    ...Object.keys(relatedPackages),
+    'react',
```
simoneb, line 21: Q: why do we need to add these manual entries? Doesn't expo include these in their "related packages"?

anapaulalemos: React and react native packages are present in the API but with different names:

Screenshot 2023-04-21 at 08 54 33

I manually declared them because I saw that's how it's done in `expo-doctor`: https://github.com/expo/expo-cli/blob/main/packages/expo-doctor/src/utils/getRemoteVersionsForSdkAsync.ts#L8

simoneb: ok thanks. are we not missing one then? :)

scripts/check-expo-dependencies.js
```diff
@@ -0,0 +1,43 @@
+module.exports = async ({ context, github, fetch, dependencyNames }) => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const expoVersion = require('../package.json').dependencies.expo
+  const expoVersionSanitized = expoVersion.replace(/[!^~]/g, '')
```
simoneb, line 4: can we make the regxp slightly better?

1. why "!"? is that a valid version specifier?
2. why the "g" modifier? we know where the version specifier can occur and how many there can be. only one at the beginning of the version, so the regexp can be more specific

scripts/check-expo-dependencies.js
anapaulalemos: My bad 🙄  Thanks for noticing

simoneb (approved):

simoneb:
lgtm. @anapaulalemos can you please keep an eye on the PRs coming in to confirm that the right dependabot PRs are being correctly closed?

Issue #1186, opened 2023-05-19
Include information in push notification about which version of the package is being published


simoneb:
At the moment the push notification doesn't include this information, we should check if it's feasible and easy to include it as it would provide more context to the person approving the OTP generation

PR #1198, opened 2023-05-31
feat: send package info when requesting OTP

guilhermelimak:
This PR adds support for displaying package information (name and version) when receiving an OTP notification both when the 
user clicks on the notification and on the notifications list.

image

image

Reference nearform/optic-expo#1186.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   "devDependencies": {
     "@babel/core": "^7.19.3",
     "@babel/preset-typescript": "^7.21.5",
-    "@commitlint/cli": "^17.6.3",
+    "@commitlint/cli": "^17.6.5",
     "@commitlint/config-conventional": "^17.6.3",
     "@react-native-community/eslint-config": "^3.1.0",
     "@testing-library/jest-native": "^5.4.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.3.10",
   "versionCode": 21,
   "scripts": {
-    "start": "expo start --dev-client",
```
simoneb, line 7: can you explain why this is not needed?

src/Main.tsx
```diff
@@ -55,6 +55,10 @@ export type MainStackParamList = {
     secretId: string
     token: string
     uniqueId: string
+    packageInfo?: {
+      name?: string
+      version?: string
+    }
```
simoneb, line 61: shall we create an issue to track making this (and changes done in other projects) mandatory some time in the future?

package.json
guilhermelimak: I'm not sure why this was here in the first place since the `--dev-client` flag should be the difference between `start` and `start:native` [according to the README](https://github.com/nearform/optic-expo#notes)

src/Main.tsx
guilhermelimak: I would say so, but is there a timeframe where it would be fine to change that? Because it would break users workflow until they update the action and since we don't receive any update notifications for action releases (and dependabot won't work with actions either) they will never know the action changed.

guilhermelimak: Actually I think we can change our current version of the action and add a warning telling the user to update until a certain date (a couple of months from now probably since this feature is not really important) when we would remove API's backwards compatibility. That way I think most people would notice and update the action since the user won't have to change anything on their side.

When we do go forward with the API update we should also add a helpful error message for the user explaining why their CI is broken and what to do to fix it.

simoneb: A better way to handle this is to release a major of the action instead, users don't look at warnings

guilhermelimak: but this won't really solve the issue I don't think. The users will have no way of knowing that we released a new major since they won't be notified and everyone on the old version will have the CI broken when we finally remove backwards compatibility

PR #1225, opened 2023-06-20
feat: add action to update expo QR codes on optic landing page

guilhermelimak:
This PR adds a new action that will run after a release and will update the Expo Go QR codes on the `optic-website` docs.

### DIFF
--- a/.github/workflows/replace-expo-qr-codes.yml
+++ b/.github/workflows/replace-expo-qr-codes.yml
@@ -1,7 +1,7 @@
 name: Replace Expo update ID in README
 
 on:
-  push:
+  workflow_call:
 
 jobs:
   replace-expo-update-id:
--- a/.github/workflows/replace-expo-qr-codes.yml
+++ b/.github/workflows/replace-expo-qr-codes.yml
@@ -1,0 +1,96 @@
+name: Replace Expo update ID in README
+
+on:
+  push:
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/replace-expo-qr-codes.yml
```diff
@@ -0,0 +1,96 @@
...
+        id: get-expo-update-id
+        run: |
+          UPDATES=$(eas update:view --json $(eas branch:view production --json --non-interactive | jq ".currentPage[0].group" -r ))
+          echo "::set-output name=android_update_id::$(echo $UPDATES | jq -r '.[] | select(.platform | contains("android")) | .id')"
+          echo "::set-output name=ios_update_id::$(echo $UPDATES | jq -r '.[] | select(.platform | contains("ios")) | .id')"
+
+      - name: Replace Expo update ID
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.PAT }}
```
simoneb, line 32: ok no we can't do this. let's move this to the website repo and run on a schedule instead

guilhermelimak: What do you think would be a good frequency for running it? since it will just exit if there's nothing to update I think it could be run relatively frequently.

simoneb: once per hour?

PR #1248, opened 2023-08-04
added production track to android

### LINKED ISSUE #1245: Can we publish directly to the "production" environment of the app stores?
This application is being automatically published to the stores via `eas submit` when there is a release made with Optic, as usual.

The problem is that it goes to a "channel", both in the Apple and Play stores, that is for internal testing and requires manual steps to take to production. Those manual steps furthermore require that somebody logs into the stores websites, which few people have access to.

We would like to see if it's possible, given the nature of the application, to skip this step altogether and, when a new release is made in the repo, publish the store apps directly to production.

Relevant docs exist on the Expo website: https://docs.expo.dev/submit/introduction/. Everything else can be found in this repository, such as the [eas configuration file](https://github.com/nearform/optic-expo/blob/master/eas.json), and the files where the expo commands are executed are the yaml workflow files.

People with some level of knowledge on this topic that you may want to reach out to are @confraria  and @manosim .

jamesthedove:
This is to ensure that android deployments are released to immediately to users without having to go through the manual process of creating a production release on the play store.
fixes #1245

### DIFF
--- a/eas.json
+++ b/eas.json
@@ -24,6 +24,9 @@
     "production": {
       "ios": {
         "ascAppId": "1643969791"
+      },
+      "android": {
+        "track": "production"
       }
     }
   }


simoneb (approved):
LGTM

Issue #1155, opened 2023-05-03
Opening app from notification when fully closed can swallow the notification


AlanSl:
I had this a couple of times on the Android app as published on Play Store. Not sure if it's Android-specific because I was having other issues with the iOS app which prevented me from testing it.

To replicate, on a real device and real app (not tested on Expo Go which might behave differently): 

 - Fully close the app on Android (swipe up from device app switcher screen)
 - Do an optic release
 - Open the app from the notification in the device status bar 

If the  app is merely in the background and the session is still logged in, this seems to work fine and open the approve or reject 2FA screen as expected.

 If it was fully closed and therefore also logged out, however, what seems to happen is the app boot redirects to log in, the log in process flashes up and completes quickly automatically as expected, but the app then goes to the home screen instead of the 2FA screen for the current notification.

Also, there isn't any red notice in the Notifications icon or a way to access the 2FA challange from the notifications window, so I guess what is probably happening is, the booting app initially opens the 2FA challenge, marks it as read, then navigates to the log in before the user can respond, then after completing log in, it sees no unread 2FAs, and navigates to the home screen?

PR #1256, opened 2023-08-15
fix: allows OTP to be approved from a cold start

williamlines:
Adds a conditional to the notification listener useEffect to use the necessary callbacks when a OTP request is pressed whilst the app is closed.

closes: https://github.com/nearform/optic-expo/issues/1155

### DIFF
--- a/src/context/InitalLoadingContext.tsx
+++ b/src/context/InitalLoadingContext.tsx
@@ -1,4 +1,10 @@
-import React, { createContext, useContext, useState } from 'react'
+import React, {
+  createContext,
+  useContext,
+  useState,
+  useMemo,
+  useCallback,
+} from 'react'
 
 const initialContext = {
   initialLoadingComplete: false,
@@ -14,14 +20,20 @@ export const useInitialLoading = () => useContext(InitialLoadingContext)
 export const InitialLoadingProvider = ({ children }) => {
   const [initialLoadingComplete, setInitialLoadingComplete] = useState(false)
 
-  const markInitialLoadingComplete = () => {
+  const markInitialLoadingComplete = useCallback(() => {
     setInitialLoadingComplete(true)
-  }
+  }, [setInitialLoadingComplete])
+
+  const value = useMemo(
+    () => ({
+      initialLoadingComplete,
+      markInitialLoadingComplete,
+    }),
+    [initialLoadingComplete, markInitialLoadingComplete]
+  )
 
   return (
-    <InitialLoadingContext.Provider
-      value={{ initialLoadingComplete, markInitialLoadingComplete }}
-    >
+    <InitialLoadingContext.Provider value={value}>
       {children}
     </InitialLoadingContext.Provider>
   )
--- a/src/context/InitalLoadingContext.tsx
+++ b/src/context/InitalLoadingContext.tsx
@@ -1,0 +1,28 @@
+import React, { createContext, useContext, useState } from 'react'
+
+const initialContext = {
+  initialLoadingComplete: false,
+  markInitialLoadingComplete: () => {
+    // set in context provider construction
+  },
+}
+
+const InitialLoadingContext = createContext(initialContext)
+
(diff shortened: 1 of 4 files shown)


williamlines:
This is currently not working on IOS

simoneb (commented):
The gut feeling when looking at this change is that there should be a more straightforward way to achieve what we want. The assumption is that we're not doing anything fancy with push notifications, we just need to open the app when we get one, open the app in a specific "state". 

Hence, I have a feeling that if it requires a solution like this one, which is moderately complex and error prone, there's a good chance we're doing something wrong.

On the other hand I don't have a solution either, but I would try to spend some time trying to understand what the root problem is rather than trying patching the issue.

williamlines:
I have done some more research for alternative solutions since learning that this approach is possibly non-viable for IOS, I have a lead for a solution that feels less hacky, by using the `NotificationResponseReceivedListener`, the docs suggest that this should be able to use the notification from a cold start, need to figure out the implementation.

Screenshot 2023-08-15 at 15 18 44

simoneb:
Good lead Will, worth investigating

williamlines:
The problem currently is that the notification is consumed to open the app, before the listeners are active, hopefully the `NotificationResponseReceivedListener` should be able to catch the notification.

simoneb:
I am not terribly familiar with how this issue manifests itself, but in the app we have a notifications section. If we managed to go in there and keep the notification visible in the list, it may be an acceptable compromise. But I'm just making this up now, let's definitely investigate what seems to be a more appropriate solution

williamlines:
In the case of both adding a notification to the list and displaying the 2FA challenge, there is the same difficulty of having to 'catch' the notification before the app starts up, or it gets lost. My feeling is that such a compromise likely won't save much effort. I will see how this listener behaves before jumping to conclusions

williamlines:
Upon investigating, it seems like this implementation is the same as what we currently have. The docs would suggest that this implementation should work for our needs but it seems like this is not the case, see: 
https://github.com/expo/expo/issues/14078
The solution many suggest in this thread is similar to what I have done here, as far as I can find there isn't a much better solution.

src/screens/HomeScreen.tsx
```diff
@@ -47,6 +47,8 @@ export const HomeScreen: React.FC<HomeScreenProps> = ({ navigation }) => {
   const isFocused = useIsFocused()
   const notificationListener = useRef<Subscription>()
   const responseListener = useRef<Subscription>()
+  const lastNotificationResponse = Notifications.useLastNotificationResponse()
+  const [initialLoadingComplete, setInitialLoadingComplete] = useState(false)
```
simoneb, line 51: one Q: if we store this state in the component and we navigate away from this screen, the next time we return to this screen the state will be reinitialized to false. Is this what we expect?

simoneb: I agree that conceptually this is wrong. On the other hand there's a chance that, e.g. because of how the app navigation works, this may not be an issue in practice. I would nevertheless look into it to either confirm that it's not an issue or find a different solution.

williamlines: resolved in 519d0d67de4d0b9a131685cc1d39419f3e0efd1e

src/context/InitalLoadingContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const InitialLoadingProvider = ({ children }) => {
+  const [initialLoadingComplete, setInitialLoadingComplete] = useState(false)
+
+  const markInitialLoadingComplete = () => {
+    setInitialLoadingComplete(true)
+  }
+
+  return (
+    <InitialLoadingContext.Provider
+      value={{ initialLoadingComplete, markInitialLoadingComplete }}
```
simoneb, line 23: prefer memoizing this object, otherwise each render of this component will create a new object, which will cause a rerender of all users of this context, which is unnecessary

simoneb (approved):
LGTM 🚀

Issue #1278, opened 2023-09-26
Set up a system to update Apple certs


williamlines:
The certificate needed to build and release on the app store has expired, we did not have an indication that this had happened until build time. We should set up a system to either automatically renew this cert or at least notify us when it expires.

PR #1279, opened 2023-09-26
docs: document release certificate generation process

williamlines:
Documenting the process for creating new certificates for releases to the iOS App Store.

closes: https://github.com/nearform/optic-expo/issues/1278


simoneb (commented):
All good for me, just making sure we're not disclosing sensitive information in the screenshots especially, right? This repo is public.

simoneb (commented):
Also one Q: why would the certificate expire in the first place? Is there a way to prevent that from happening?

williamlines:
@simoneb The max duration you can specify for a certificate is one year, unfortunately. So this will need to be an ongoing consideration

williamlines:
> All good for me, just making sure we're not disclosing sensitive information in the screenshots especially, right? This repo is public.

I think this is safe but I will double check with Manos

williamlines:
The screenshot is safe to share

simoneb (approved):
LGTM, thanks @williamlines and @manosim for looking into it

PR #1350, opened 2024-05-22
Fix: added missing workspace permission

synapse:


simoneb (commented):
This change is not actually necessary. id-token: write is only necessary when using provenance, meaning only when publishing an npm package.

Issue #1388, opened 2025-01-20
Post SDK 52 migration steps


simoneb:
- [ ] can we keep the "deploy previews" feature? [the action](https://github.com/expo/expo-preview-action) that we were using (and currently working around [here](https://github.com/nearform/optic-expo/blob/master/.github/workflows/expo-eas-update.yml#L41)) is now deprecated in favor of [this new action](https://github.com/expo/expo-github-action), which we should check out to see if it supports the deploy/dev previews and even with publishing in the first place (because it seems to be doing more than just dev builds). NOTE: it looks like we are actually already using the new action, but we weren't using it for the preview build, possibly because this feature for formerly into the now-deprecated action. To be looked into!
- [ ] publishing to app stores: I would suggest that we remove that part altogether because keeping the accounts active and fully automating that step is more effort than it's worth, therefore let's remove any code that tries to automatically publish to app stores
- rebrand: please check out the updated designs in figma (link in the readme) and align the app styles to the new designs
- [ ] consider switching to new architecture
- [ ] try out the new expo file system API
- [ ] let's try the app on Apple devices as well (let's ask somebody internally if they have a device to try it on)

simoneb:
I added another item for the website, which displays QR codes for downloading the app. We need to address that too.

simoneb:
@finnmerlett updated the website QR codes item above to clarify the need

PR #1396, opened 2025-01-27
Post SDK 52 migration | Step 2: disable app stores cd release step

finnmerlett:
Closes item 2 in the post-migration steps

Relates to #1388


simoneb (commented):
We can remove this for good, if we're ever going back to publishing to app stores automatically we're redo it from scratch (especially as it will have most likely changed by then)

simoneb:
Also as discussed, let's also do a related change to the optic website to remove any references to the app stores

